PR #31, opened 2023-01-09
Event deletion fix + silence a couple of warnings

msub2:
- Fixes being unable to delete events from the database. Prisma expects a number for the id but it was getting a string
- Fix an empty `for` attribute on a label causing console spam whenever you moused over the playtest input field
- Replace hex color string with a hex literal


lalalune:
@all-contributors please add @msub2 for code

Issue #6, opened 2023-01-07
Add import to main menu / move to top level menu


lalalune:
Right now you have to go to "open" to import, import should just be in the file menu

PR #45, opened 2023-01-09
Add import shortcut to File menu

msub2:
Addresses #6. A lot of the actual functionality is tied up inside functional React components so I've opted to simply condense going to the all-projects page and clicking import into a single action.


michaelsharpe (approved):
Looks good to me!  It would be good to eventually make a modal that has a file drop area on it for import, but we can do that down the road.  This works great to get it popping up as soon as you click it.

Issue #15, opened 2023-01-08
Jupyter Notebook Integration


lalalune:
Allow users to call a running Notebook from their Jupyter server by name (ideally validated via dropdown menu)

Bounty here
https://app.dework.xyz/m3/ai-43567?taskId=7b7d8230-9e9f-4add-a090-06781768cf30

lalalune:
@all-contributors please add @sshivaditya2019 for code

PR #46, opened 2023-01-09
Added the Jupyter Node

sshivaditya2019:
Adds the Functionality mentioned in #15. 

image

The output of the node would be the print output from the cell.


sshivaditya2019:
The following steps need to be done before using the Jupyter Server

- jupyter notebook --generate-config
- In Linux the configration would be at `~/.jupyter/jupyter_notebook_config.py` in windows the same would be at `C:\Users\USERNAME\.jupyter\jupyter_notebook_config.py`
- Add the Following lines in the jupyter config file
```
c.NotebookApp.ip = '0.0.0.0' # listen on all IPs
c.NotebookApp.token = ''     # disable authentication
c.NotebookApp.allow_origin = '*' # allow access from anywhere
c.NotebookApp.disable_check_xsrf = True # allow cross-site requests
```

The execute command at the front end can access the code only from the last cell of the jupyter notebook, so if functions are to be used they have to be used in the same cell.

lalalune (approved):
I haven't tried this with a complex notebook yet but it looks good to me!

PR #47, opened 2023-01-09
Switch to whisper

hozzjss:
I'm keeping this PR open during the development of the feature for incremental reviews and updates


lalalune:
Fantastic, let me know if you need any help!

lalalune:
Looks like it's coming along. Just fixed a merge conflict with the .env

hozzjss:
Hey @lalalune I think this is gonna take more time and a special version of the library, the stream thingy isn’t wrapped correctly the implementation isn’t intended for a client/server model so might need to either fuck with the cpp or just keep it in the client kinda sucks, will need time for server to take charge of it

lalalune:
@hozzjss what is your confidence in this approach? Can raise the bounty given the complexity

hozzjss:
The short answer is that the whisper.cpp implementation looks like a PoC rather than a finished lib, the problem is that it was not built to scale or handle multiple user requests rather to play with or for simple client based apps

hozzjss:
I’m not a 100% sure that it’s that complicated objectively as i am certain that for someone with more knowledge of cpp and integration with nodejs would have an edge over me in estimating how complex this is, i would still prefer the client based model but for ux i might just test the user’s hardware first or make it optional as the requests to Google will be slower if someone has a capable pc that can process the whisper wasm without breaking the ui or frying the cpu lol

hozzjss:
I'm going to close this PR since I won't be able to put any effort into it atm with my lack of knowledge about c++, someone else could pick up where I stopped

PR #57, opened 2023-01-14
fixed process code, and python node

khalildh:


lalalune:
@all-contibutors please add @khalildh for code

lalalune:
@allcontributors please add @khalildh for code

PR #66, opened 2023-01-17
Added Question Answering Node

sshivaditya2019:
Added Question Node for Weaviate Neural Search. It can answer questions based on conversations stored as events.

### DIFF
--- a/packages/core/types.ts
+++ b/packages/core/types.ts
@@ -52,6 +52,19 @@ export type Event = {
   date?: string
 }
 
+export type SemanticSearch = {
+  concept?: string
+  postive?: string
+  negative?: string
+  distance?: number
+  postive_distance?: number
+  negative_distance?: number
+}
+
+export type QAArgs = {
+  question: string
+}
+
 export type CreateEventArgs = Event
 
 export type GetEventArgs = Event & {
@@ -125,6 +138,7 @@ export type EngineContext = {
   ) => Promise<string | string[] | null | Record<string, any>>
   storeEvent: (args: CreateEventArgs) => Promise<any>
   storeEventWeaviate: (args: CreateEventArgs) => Promise<any>
+  eventQAWeaviate: (args: QAArgs) => Promise<string | string[] >
   getWikipediaSummary: (keyword: string) => Promise<Record<string, any> | null>
 }
 
--- a/packages/core/types.ts
+++ b/packages/core/types.ts
@@ -54,10 +54,10 @@ export type Event = {
 
 export type SemanticSearch = {
   concept?: string
-  postive?: string
+  positive?: string
   negative?: string
   distance?: number
-  postive_distance?: number
+  positive_distance?: number
   negative_distance?: number
 }
 
--- a/packages/core/types.ts
+++ b/packages/core/types.ts
@@ -63,6 +63,7 @@ export type SemanticSearch = {
 
 export type QAArgs = {
   question: string
+  agentId: string
 }
 
 export type CreateEventArgs = Event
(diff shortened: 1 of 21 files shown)


packages/systems/src/systems/weaviate_events_schema.ts
```diff
@@ -17,50 +23,62 @@ export default
...
+        "description": "Name of the channel through which event was created.",
+        "name": "channel"
       },
       {
         "dataType": [
           "string"
         ],
-        "description": "Contains the output of the text embedded in the text",
-        "name": "text"
+        "description": "Type of the channel thourgh which event was created.",
```
vincentfretin, line 61: thourgh -> through

packages/systems/src/systems/weaviate_events_schema.ts
```diff
@@ -17,50 +23,62 @@ export default
...
           "string"
         ],
-        "description": "The date at which event was created",
+        "description": "The date at which the event was created",
         "name": "date"
+      },{
+        "dataType": [
+          "string"
+        ],
+        "description" : "The Observerver",
```
vincentfretin, line 80: Observerver -> Observer

packages/systems/src/systems/weaviateInit.ts
```diff
@@ -161,4 +151,78 @@ export class weaviate_connection {
...
+                                        .explore()
+                                        .withNearText({
+                                          concepts: [data["concept"]],
+                                          distance: 0.6,
+                                          moveAwayFrom: {
+                                            concepts: [data["negative"]],
+                                            force: data["negative_distance"]
+                                          },
+                                          moveTo: {
+                                            concepts: [data["postive"]],
```
vincentfretin, line 212: postive -> positive

packages/systems/src/systems/weaviateInit.ts
```diff
@@ -161,4 +151,78 @@ export class weaviate_connection {
...
+    }
+    
+  }
+
+  static async semanticSearch(_data: SemanticSearch){
+    if (!weaviate_client) {
+        await initWeaviateClientEvent()
+    }
+    const data = { ..._data }
+    const validFields = ['concept', 'postive', 'negative', 'distance', 'postive_distance', 'negative_distance']
```
vincentfretin, line 196: postive -> positive
postive_distance -> positive_distance

packages/core/types.ts
```diff
@@ -52,6 +52,19 @@ export type Event = {
   date?: string
 }
 
+export type SemanticSearch = {
+  concept?: string
+  postive?: string
```
vincentfretin, line 57: postive -> positive

packages/core/types.ts
```diff
@@ -52,6 +52,19 @@ export type Event = {
...
+  negative?: string
+  distance?: number
+  postive_distance?: number
```
vincentfretin, line 60: postive_distance -> positive_distance

lalalune:
@sshivaditya had a chance to review

- Storing events works well.

- Unable to recall events using the event recall node

- QA works well

Two things are needed -- a way to delete events and the ability to recall and filter by the entities.

The problem is that when the agent speakers, the sender and observer are both agent. To keep track of which conversation we are in, we need to key it by the entities -- speaker and agent at the very least, but there could be 3 speakers and an agent. We'd want the last conversation where the entities include both speaker and agent, whether other people came or left.

We are doing this conversation in a 3D world where other users and agents might walk up, so tracking just 2 speakers in the convo doesn't make sense.

We also have multiple agents, so when we ask the question, we need to make sure it's keyed to only recall the information the current agent knows.

sshivaditya2019:
> Unable to recall events using the event recall node

Recall Node was returning the events in Array format. Fixed by converting into JSON and returning, the same.

> Two things are needed -- a way to delete events and the ability to recall and filter by the entities.

Added a Delete Node, needs all values same as the store Node. These values are used to get the ID of all matching events, which are then deleted.

Now, the event recall will filter events based on observer and sender, only those conversations where the sender, observer and agent were active will be returned.

> We also have multiple agents, so when we ask the question, we need to make sure it's keyed to only recall the information the current agent knows

Added an agent id based filter for the question answer node.

PR #71, opened 2023-01-20
Big refactor: Add Feathers, Consolidate Server and Packages

lalalune:
This adds Feathers 5, which also gives us an opportunity to consolidate and drastically simplify our architecture.

- Remove Prisma - Feathers 5 natively interates Typebox and converting was easy
- Remove separate servers - One server that starts up all backend services
- Consolidate core and engine packages
- Remove Magick interface - was doubling up a lot of code and now that we are moving server-only with nodes, unnecessary
- Remove gamestate - we will register and populate private and public variables directly on the graph, adds a lot of weight


lalalune:
@michaelsharpe noticed that if i change a spell or save, its not updated.

instead of dirtying agents and spells and stuff, lets have the entity add itself to the db as an active connection, and then send out messages to all running agents and servers to reload the spell? think this would be a cleaner and more scalable pattern

PR #84, opened 2023-01-31
Feature: Fine Tune Manager

thomageanderson:
Deliverables from [Dework Task:
](https://app.dework.xyz/magickml?taskId=70b04ebb-fccb-43c4-8659-7e62acef8487)
- [x] Adapt code from [fine-tune](https://github.com/assaf/fine-tune) into the Magick stack
- [x]  Embed into a “Fine Tune Manager” drawer window in our main client
- [x]  Make OpenAI fine tunes accessible in the completion node for live use
- [ ] Models should be tracked in a database
- [ ] Model endpoint should make them available from a node dropdown to select
- [ ] Vendor flexibility to allow for future provider integration
- [ ] Clean up styles and check that we are using comparable MUI components

Questions:
- Do we want to support OpenAI classification and search models/endpoints? If so, should that immediately change the Inspector layout to have a nested select for models by task/type ? Or are these just in different rete nodes to separate concerns?

- Are there other example of inspector components that are receiving dynamic data from the React layer? I've temporarily hacked the dropdown select to make the list endpoint network call but wasn't entirely sure where that should be managed. I'm assuming a vendor accounts context that deals with credentials and available resources from each.

- Is tracking the models in a database duplicating work when we're listing directly from the vendor anyway? Seems like it introduces an issue of being out of synch for instance

- Similar concern regarding storing user specific vendor credentials in some kind of settings table - if we do that is some kind of CRUD management panel for vendor credentials in scope for this task? Personally I think the local storage pattern cuts out a lot of complexity here while deliverable similar immediate value

- Regarding The Drawer Component: Could we consider calculating the current selected tab based on the URL i.e `/magick` ,  `/fineTuneManager`


lalalune:
Questions:

Do we want to support OpenAI classification and search models/endpoints? If so, should that immediately change the Inspector layout to have a nested select for models by task/type ? Or are these just in different rete nodes to separate concerns?
For classification and search we are really targeting Weaviate, since OpenAI's search is just so expensive. I suspect that we don't want to deal with those unless there's a specific ask. Most people don't even know OpenAI does search and just want fine tuned generations or embeddings.
Ideally we'd have a way to create the embeddings and load them into Weaviate but that sounds out of scope for now.

Are there other example of inspector components that are receiving dynamic data from the React layer? I've temporarily hacked the dropdown select to make the list endpoint network call but wasn't entirely sure where that should be managed. I'm assuming a vendor accounts context that deals with credentials and available resources from each.
Yes, thje Spell component.

Is tracking the models in a database duplicating work when we're listing directly from the vendor anyway? Seems like it introduces an issue of being out of synch for instance
I think it is, and we want to minimze unnecessary DB storage to keep the backend light.

Similar concern regarding storing user specific vendor credentials in some kind of settings table - if we do that is some kind of CRUD management panel for vendor credentials in scope for this task? Personally I think the local storage pattern cuts out a lot of complexity here while deliverable similar immediate value
I like local storage and agree. I think that long term we need a full users settings in our user table to store various keys and such that we'lls need to access, but at the same time we're trying to just abstract all of that with a Magick account so you don't need to worry about API keys. The current method has the keys loaded into the Agent when that is saved.
For now, the simplest option with the least coding is the best one.

Regarding The Drawer Component: Could we consider calculating the current selected tab based on the URL i.e /magick , /fineTuneManager
Yes I think we could do that in the router, sure

FabriceIRANKUNDA:
so far LGTM!
At the usage instructions section, the link to more details can be made a href to make it easy for the users to navigate there by one click.

michaelsharpe:
One quick comment here, if it is in scope.  It would be amazing if the fine-tuning panel showed datasets generated by the nodes themselves running, so you could easily take any node and fine-tune a smaller model on its collected data.

lalalune:
> One quick comment here, if it is in scope. It would be amazing if the fine-tuning panel showed datasets generated by the nodes themselves running, so you could easily take any node and fine-tune a smaller model on its collected data.

Cool but IMO we should split that to be a separate PR.

Issue #52, opened 2023-01-12
Typing "Shift+F" will create a comment box on the graph even if you don't have it focused. Same with Shift+C


msub2:
For example, naming a node "Agent Facts" will create a comment box as I'm typing

PR #85, opened 2023-02-01
Fix comment box creation

FabriceIRANKUNDA:
#52


packages/client-core/src/plugins/commentPlugin/index.js
```diff
@@ -23,7 +23,7 @@ function install(
       altKey: false,
     },
     deleteCommentKeys = {
-      code: 'Delete',
+      code: 'Backspace',
```
lalalune, line 26: We actually want both delete *and* backspace

packages/client-core/src/plugins/commentPlugin/index.js
```diff
@@ -57,7 +57,9 @@ function install(
         e.code === x.code &&
         e.shiftKey === x.shiftKey &&
         e.ctrlKey === x.ctrlKey &&
-        e.altKey === x.altKey
+        e.altKey === x.altKey &&
```
lalalune, line 60: LGTM

Issue #78, opened 2023-01-26
Delete does not get registered 


chintanpuggalok:
The delete/backspace does not remove the nodes

lalalune:
Great, we should add backspace to delete nodes, instead of the delete

PR #87, opened 2023-02-02
Fix delete event register

FabriceIRANKUNDA:
#78


michaelsharpe:
Great work!
One thing to note here. We generally want to keep the lock file as it can contain important exact matches to dependencies. In theory the lock file should always represent a working state of all dependencies.

FabriceIRANKUNDA:
> Great work! One thing to note here. We generally want to keep the lock file as it can contain important exact matches to dependencies. In theory the lock file should always represent a working state of all dependencies.

Alright! I'll keep that in mind

lalalune (approved):

PR #93, opened 2023-02-05
Adds 3OV for WordPress template

antpb:
For your consideration :) 

This PR adds a 3OV for WordPress template that helps people get started quicker with the 3OV plugin and Magick. This template is primed to receive data from a WordPress install using the NPC block.

createnew


lalalune:
lgtm

Issue #99, opened 2023-02-06
If a component has a name, show it


lalalune:
Some components can be named - for example variables

We removed this functionality because it was changing the name of the actual component and breaking things.

What we want is to show the node name on the node, if it exists

This is the current
![image](https://user-images.githubusercontent.com/18633264/216893614-7a43ddc3-7916-410b-8d53-1eb333aa01f0.png)

We want more like this:
![image](https://user-images.githubusercontent.com/18633264/216893955-d29703bc-04a4-457d-95b5-4b37ab4de3e2.png)

PR #106, opened 2023-02-08
Ft component issue

FabriceIRANKUNDA:
#99 
[x] Fixed showing component name when exist


lalalune:
Handled in separate PR, was a one-liner

PR #118, opened 2023-02-11
fix(feathers-api): ui send hash of spell graph nodes

alexjorgef:
When creating a new spell send a hash of graph nodes, this is required for API validation


lalalune:
@allcontributors please add @alexjorgef for code

PR #136, opened 2023-02-12
Fine Tune Manager: UI and Vendor Context Updates

thomageanderson:
Known outstanding issues:

- [ ] jsonl file upload format isn't working ( bug present in the fine-tune source we referenced / ported)
- [ ] JSON markdown snippets have an overflow issue
- [ ] react-select could be replaced or updated to implement MUI using the react-select Controller component
- [ ] would benefit from a scroll view in some cases, when generate test completions for instance on the fine tune testing page


lalalune:
@thomageanderson i have repointed this at the v0.1.0 branch -- i had merged the work from before to keep it current and roughly tested the interface

thomageanderson:
thanks @lalalune , I walked back my merge from main and resolved conflicts with `v0.1.0` in that case

PR #138, opened 2023-02-13
Request for Similar Embedding Works, with content query param

sshivaditya2019:


packages/engine/src/nodes/events/EventSearch.ts
```diff
@@ -0,0 +1,87 @@
...
+  MagickWorkerInputs,
+  MagickWorkerOutputs,
+} from '../../types'
+import { InputControl } from '../../dataControls/InputControl'
+import { triggerSocket, stringSocket, eventSocket, arraySocket, anySocket } from '../../sockets'
+import { MagickComponent } from '../../magick-component'
+import { API_ROOT_URL } from '../../config'
+
+
+const info = 'Event Search is to find simlar Events '
```
lalalune, line 18: similar

packages/plugin-discord/src/components/agent.component.tsx
```diff
@@ -7,7 +7,7 @@ type PluginProps = {
   agentData: any
   props
 }
-import {API_ROOT_URL} from '@magickml/engine'
+import {API_ROOT_URL} from '../../../engine/src'
```
lalalune, line 10: we shouldnt need this, might be an issue with our routes

lalalune:
@sshivaditya looks good but we can probably consolidate this into the "event recall" node and only perform the search if the embedding field is not null or not null vector (array dim 1536 of all 0s)

packages/plugin-discord/src/components/agent.component.tsx
sshivaditya2019: Adding `@magickml/engine` causes circular dependency. The plugin.ts from engine package depends on Plugin_Discord which would now depend on `@magickml/engine`, if this is added, causing circular dependency

sshivaditya2019:
@lalalune Merged with EventRecall, both are serviced using an after hook from Event Service.

packages/plugin-discord/src/components/agent.component.tsx
lalalune: Dang that's so annoying :P

PR #158, opened 2023-02-17
[WIP] Build Editor Package with Types

lalalune:
This is a WIP. Currently package builds and publishes but the types are not importing properly into a separate next.js project. Many type errors are thrown on the type generation at the end.


lalalune:
- Some styling got lost (probably index.css ref or something)
- Need to pass more env vars in via config object (basically all client side ones)
- Need to merge and handle conflicts

Screenshot 2023-02-17 at 9 11 14 PM

PR #161, opened 2023-02-17
151 subspells broken

michaelsharpe:
Should be tested. Please ensure that:

- you can select an existing spell from the spell component dropdown
- you can create a new spell form the spell component dropdown
- The sockets in and out appear properly
- input and output sockets are updated when the subspell IO changes in another tab
- spell component will run the subspell in the IDE and on the entity server


michaelsharpe:
Almost done.  Weird issue with the runSpell function not managing to access the subspell in the database, likely due to an improper match against projectId.

lalalune:
This has a known UI bug, but it functionally works, so lfg

PR #192, opened 2023-02-20
Development override openai endpoint

lalalune:


michaelsharpe:
@lalalune Is this just for OpenAI? Or will this work against any other API we want to wrap as well?

Issue #41, opened 2023-01-09
Public variables


michaelsharpe:
Public variable nodes. Should have different types. Could possibly just make our current variables exposable as public.  A public variable can be inject into the spell. So either as data from outside or as sockets into a spell. Used to create sets of data for different entities to share the same brain/spell. 

Need to consider whether public variables can be populated from the top level spell all the way down to subspells or not.

lalalune:
@FabriceIRANKUNDA next step on this

- all public variables for the "root spell" chosen are exposed in the agent's dropdown
- we pass this data into the spell runner when we run the spell (or when we load it?)
- goal is a bot that says "hi my name is <name" with the name coming from a public variable that is set on the agent

lalalune:
https://user-images.githubusercontent.com/18633264/219939675-7205b989-a628-40da-a2dd-b61295648ac3.mp4

lalalune:
![Group 1](https://user-images.githubusercontent.com/18633264/219939699-2d2c31b5-f517-4027-b39e-7d2aaa801145.png)

michaelsharpe:
One thing I would like to see is a set component for each variable type. Prompt input is a text area. Array input would be some kind of tag list perhaps (though arrays are more complicated if they involve complex objects). Boolean would be a switch. A number could be a range slider or just a number text area. String is just an input field. Etc. 

These will be programmatically composed from a set of react components. This will make it easier to style these as well.

FabriceIRANKUNDA:
> 

Exactly! basically experienced that challenge when displaying variables and having a component for each would make the rendering much easier. I think I can set the component to each variable.

PR #202, opened 2023-02-22
display public variables of selected spell in the agent

FabriceIRANKUNDA:
#41


lalalune:
Variable looks good... now we want to make it editable.

The next step will be to make the variables public and save the data onto the agent.

We will need to make a new field on the agent for agentPublicInputs to store these on.

Screenshot 2023-02-21 at 6 34 46 PM

lalalune:
Looking good, a few bugs

- If I only have one spell, and i load the agent, the public vars never show
- If I save the public vars it wipes the rest of my agent

PR #203, opened 2023-02-22
Fixed Z-Index of Inspect Window's select elements 

matei3d:
There was an issue with the z-index of one of the Inspect Window's inspect elements. 

It was causing it to draw on top of top File dropdown, blocking the Save button.

Before: 
select_borked

After:
select_fixed


lalalune:
@all-contributors please add @matei3d for code

PR #238, opened 2023-02-24
ID based key for spells

### LINKED ISSUE #186: Spells should have IDs, use ID as key instead of name
Right now if two users with different project IDs on the same database try to make a spell with the same name it will fail.

- Spells should have a UUID identifier which is their key. Each spell's UUID is unique.
- There should only be one spell with any given name for each projectId.
- There can be many spells with different names with the same projectId

### LINKED ISSUE #223: Rename Subspell Broken
The patch is attempting to save the spell by spell-name, however it needs to be saved by the spell's ID instead and sent along with the config's projectId.

Screenshot 2023-02-22 at 9 24 42 PM

sshivaditya2019:
- Updated the schema to use UUID as the primary key,
- Created URI using UUID and Name, after creating new.
- The `tab.id` is now same as UUID of the active spell
- `tab.name` will have the URI
- LocalState is filtered based on the UUID now.

Fixes #186 

- Rename is also fixed (Possibly Fixes #223 )

### DIFF
--- a/.env
+++ b/.env
@@ -1,6 +1,6 @@
 # Default settings, override in .env.local file that is auto-generated on install
 # NOTE: The default database is provided to get you up and running but is not secure!
-DATABASE_URL=postgresql://postgres:ZTE*meq1mzh3abn!cmk@db.xnzvmluhwpbngdufsbrd.supabase.co:5432/postgres
+DATABASE_URL=postgres://postgres:ZTE*meq1mzh3abn!cmk@db.xbolptvmhqplniydcjcz.supabase.co:5432/postgres
 PLUGINS=@magickml/plugin-discord
 
 # if you want to override OpenAI's endpoint with another URL that matches the same API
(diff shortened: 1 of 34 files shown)


lalalune:
Looks very close, a few notes
- spellName and name are confused, maybe we rename spellName to spellId everywhere?
- "Rename spell" - should pass the existing name as default
- "Save as" spell has the spell's UUID in the form field and doesn't work

sshivaditya2019:
Potential Issues

- `spell-runner` not updating with latest spell-name,  does not affect execution.
- Error in Engine Context, Spellref does not point to the correct active spell. (Observed only after renaming)

sshivaditya2019:
Changes

- spellName has been removed
- URI is read from the route
- URI is stored in `tab.URI`
- URI is UUID + Base64Encoded(Spell Name)

lalalune:
> Potential Issues
> spell-runner not updating with latest spell-name, does not affect execution.
> Error in Engine Context, Spellref does not point to the correct active spell. (Observed only after renaming)

Hmm, yeah, maybe @michaelsharpe can take a look at the new branch for this, `spell-uuid`

Merged with diff save, that's good.

Rename and save as work. Save as errors after performing the operation and updating.

Importing is not working.

PR #242, opened 2023-02-25
Input UI - Input has "input type" dropdown

lalalune:
This adds some UI functionality to consolidate inputs into the Input component with a dropdown

There are some bugs to look at
- [ ] Default case has trigger in and data in
- [ ] Add 'Trigger In' as input type - with custom name - only trigger, no data
- [ ] Add 'Data In' as input type - with custom name - only data, no trigger
- [ ] For Discord inputs, input type is event
- [ ] Validate that inputs for voice, text and default are all working
- [ ] Error if playtest is sending data to non-existent input (or only allow inputs which are available)


lalalune:

PR #243, opened 2023-02-25
migrate code to server-core and add rest plugin

lalalune:


lalalune:
Right now our REST is open, exposing everything.

Each agent should have a REST client option which can be enabled
Each agent has a token for the REST which can be handed to other users, and whitelisted domains
If REST is enabled, only expose the endpoint for a spell if the key is provided and the spell is enabled
Don't expose agents via REST unless authenticated

lalalune:

PR #250, opened 2023-02-27
feat(package): enable package multi-router

developerfred:


FabriceIRANKUNDA (commented):
LGMT

lalalune:
I tried running it and got `process is not defined`

My sense is that we should export all the providers as single export, all the routes as separate exports, and then make a new next.js app which exports these and builds up using it's page routing system.

If this truly works then ok, but it's adding next.js dependency to the whole component library, where what we really want is to make the component library granular enough that the components can be mapped to the next.js default way of doing things.

developerfred:
> I tried running it and got `process is not defined`
> 
> My sense is that we should export all the providers as single export, all the routes as separate exports, and then make a new next.js app which exports these and builds up using it's page routing system.
> 
> If this truly works then ok, but it's adding next.js dependency to the whole component library, where what we really want is to make the component library granular enough that the components can be mapped to the next.js default way of doing things.

I modified next is no longer needed in the project and support will still exist for other versions of routes

lalalune:
Is there an example of this working in Next.js somewhere? The code looks fine but I don't understand how to test.

developerfred:
@lalalune I will add a test in next here.

lalalune:
Closing this PR because it is merged into development but needs testing

Issue #278, opened 2023-03-03
Add scroll to dropdowns


lalalune:
Right now our dropdowns are cutting off part of the options

![image](https://user-images.githubusercontent.com/18633264/222626549-a47d5344-e270-4bd5-91f2-334a6a35069f.png)

Add a scroll to each dropdown so that users can scroll up and down on long dropdowns.

Issue #215, opened 2023-02-23
[BUG] Subspell dropdown


michaelsharpe:
if you try typing a new spell into the dropdown of the spell component in the inspector, the text is blocked out by something.  

Expected behaviour: I can type on the dropdown, see what I type, and then create and open the new spell.

Related to 151, and on the 151 spell branch

lalalune:
If I click away from the spell and back it appears messed up

<img width="685" alt="Screenshot 2023-03-03 at 4 22 08 AM" src="https://user-images.githubusercontent.com/18633264/222719062-e592825c-fa4f-402e-8c1f-5049a12ee671.png">

PR #294, opened 2023-03-03
fixed subspell bug in inspector and added scroll to dropdown component

FabriceIRANKUNDA:
#278 
#215


FabriceIRANKUNDA:
Screenshot 2023-03-03 at 15 45 43

lalalune (approved):

Issue #274, opened 2023-03-03
Close spell if it doesn't exist


lalalune:
For various reasons a spell can stop existing, especially in development.

Instead of a null graph and error, close the window or don't open it in the first place.

Issue #293, opened 2023-03-03
Throw toast if spell with same name already exists


lalalune:
If I make a spell "testing", then I make another one, nothing happens. I should get a toast from the snackbar that says that a spell with that name already exists.

<img width="601" alt="Screenshot 2023-03-03 at 3 56 13 AM" src="https://user-images.githubusercontent.com/18633264/222714439-c40d55f2-b206-426c-b67e-3fef87141dec.png">

sshivaditya2019:
Fixed in #303

Issue #292, opened 2023-03-03
Move agent name edit to details page


lalalune:
Right now the edit is in the list. We should move the edit button to the details page.

Also, it should be styled to not change from current edit style. The input should be styled to be same as header, other than the cursor to type and the edit button becoming the accept/cancel

<img width="389" alt="Screenshot 2023-03-03 at 3 24 00 AM" src="https://user-images.githubusercontent.com/18633264/222708232-825be19a-4e65-4d08-a101-4c206bb02255.png">

sshivaditya2019:
Fixed in #303

Issue #275, opened 2023-03-03
Open "create spell" if no spells exist in the db for the project yet


lalalune:

sshivaditya2019:
Fixed in #303

PR #303, opened 2023-03-04
Fix: Toast error in CreateNew for duplicates, Agent UI Changes, Tab on spell deletion

sshivaditya2019:
- Toast error in CreateNew for duplicates
- Moved agent rename into AgentDetail
- Return to Home, when a spell does not exist


sshivaditya2019:
Fixes #274 , #293, #292, #275

PR #333, opened 2023-03-09
[Plugins] Request to integrate MultiCopyPlugin

gadillacer:
Main Objectives: Copy/Paste multiple nodes with connections
Features:
- Working with built-in 'multiselectnode' event
- Add changes to PubSub Providers, menuBar, contextMenu
- Working with undo/redo plugin
- Can paste between separate graphs and keep the node connections
- Data stored in localStorage with keys 'selectedNodes' and 'connectedNodePairs'


gadillacer:
https://user-images.githubusercontent.com/107874128/224012475-921705c0-b75d-436d-975b-8f24932f7ec6.mp4

gadillacer:
https://user-images.githubusercontent.com/107874128/224016835-17665207-2b02-4104-bf55-0188e462b279.mp4

michaelsharpe:
I am not sure if you saw, but there is also a feature where you can control click to select multiple nodes.  Would this work with that feature? So you can select multiple nodes and copy/paste?

lalalune:
> I am not sure if you saw, but there is also a feature where you can control click to select multiple nodes. Would this work with that feature? So you can select multiple nodes and copy/paste?

Yes I believe that is working. Box selection is also available, but we need to change the node panning to the other mouse or something.

gadillacer:
> I am not sure if you saw, but there is also a feature where you can control click to select multiple nodes. Would this work with that feature? So you can select multiple nodes and copy/paste?

Yes, this is indeed it. control+click is Rete.js built-in 'multiselectnode' event. I just make the expansion plugin for that.

gadillacer:
Also, you can press control+c/control+v it will do the jobs.

gadillacer:
> > I am not sure if you saw, but there is also a feature where you can control click to select multiple nodes. Would this work with that feature? So you can select multiple nodes and copy/paste?
> 
> Yes I believe that is working. Box selection is also available, but we need to change the node panning to the other mouse or something.

Let's see if we can do something about that. Is that selectionPlugin?

lalalune:
LGTM, merging this for now, let's discuss multi select in next

Issue #139, opened 2023-02-13
Autosave text on click away


lalalune:
Text field / prompt not autosaving when I leave

Issue #129, opened 2023-02-12
Integrate Next auth


lalalune:
We want to integrate this into our cloud platform. To do that, we need to share auth using cookies. We have Next auth integrated into our platform, so we want to integrate it to pass auth. For the local non-platform user, auth should default to something simple and security (token, or user+password) that *just works* when starting up.

- Add auth provider hook for Magick IDE, if auth then send auth on routes
- On services add auth if auth flag is enabled
- Make sure all routes are secure and authenticated
- Integrate with Next.js platform app, either injecting auth with wrapped react component or passing a cookie to a subdomain where IDE is running

michaelsharpe:
@lalalune I will also write up a bit more here about our approach, along with a diagram for it.

I was also thinking when we pass the user object via the auth flow, we should also attach to that the users permissions, as well as their subscription status.  Will help not just authenticating, but also validating that they have the right permissions to take actions on the API

mrmetaverse:
Need IDE to be talking to Next

lalalune:
@FabriceIRANKUNDA if we are running on the cloud, we need all of our backend routes to be authorized.

@coffeeorgreentea has set up a token injection from the cloud for the JWT. We need to take this JWT from him, then pass it into the client.

@coffeeorgreentea can you please add the NEXT_AUTH_SECRET to the heroku backend instance so that it is available there

@FabriceIRANKUNDA please load the cloud and server, set the JWT secret to be the same, and lets add auth to the app.

For cloud development, the token will be provided by the cloud to the client, then validated on the server. The server and Next.js-based cloud both share the same JWT secret and decode the same tokens. The client just passes the token between them.

For local development, we should generate a hardcoded auth token locally so the user never has to worry about this, but all routes are protected.

All routes on the backend will be auth-locked down. The only one that are available will be the /rest-api which has it's own token system.

coffeeorgreentea:
> @coffeeorgreentea can you please add the NEXT_AUTH_SECRET to the heroku backend instance so that it is available there
@lalalune need 2fa to get into heroku

michaelsharpe:
One thing to note here too. The project opened per ide tab by the cloud is enclosed the the JWT. So we can validate the JWT be decoding with same secret. And then we can make sure the project the user is accessing is the right one the JWT token encodes. So the magic server won't need to constantly check with the cloud server.

PR #359, opened 2023-03-11
319 refactor websocket spell updates

michaelsharpe:
This PR resolves:
- #139 refactor websockets and fix socket connection
- #129 integrate next auth flow


michaelsharpe:
This PR still needs a few things.  SpellRunner service names should be made more clear (run instead of post, load instead of get).  Need to test this with in single user mode to make sure auth doesn't block that flow.  Right now it is probably made for JWTs only and will need to bypass auth.  Might be as simple as modifying the global auth hooks.

Issue #334, opened 2023-03-09
Enable / Disable snapping


lalalune:
Snapping is in the area plugin, just needs to be enabled/disabled

![image](https://user-images.githubusercontent.com/18633264/224157735-e545a016-ee79-440a-8ad9-0890fe34a020.png)

lalalune:
We just need to add "enable/design snap" as a menu option and have it function. By default snap should be on, I think?

PR #397, opened 2023-03-13
fix: snap enabled by default

BMO-tech:
Per #334 This is a small PR that enables `snap` by default. Quick and easy win!


lalalune:
We should add an Edit menu option for snap enable/disable - I think we should not enable it by default because it feels smoother without it, although practically OCD people will want it. We might need a snap resolution.

What i'd like more than an edit menu is some tool options in the workspace header, with snap and snap resolution there
Screenshot 2023-03-12 at 9 28 47 PM

BMO-tech:
I was just looking at that as well as a key binding for it

PR #481, opened 2023-03-21
fix: check for auth header before extracting value

BMO-tech:
This small change should resolve the inconsistent `unhandledRejection TypeError: Cannot read properties of undefined (reading 'split') ` error we've been seeing.

### DIFF
--- a/packages/server-core/src/sockets/sockets.ts
+++ b/packages/server-core/src/sockets/sockets.ts
@@ -21,13 +21,23 @@ const handleSockets = (app: any) => {
       } else {
         // todo wound up using a custom header here for the handshake.
         // Using the standard authorization header was causing issues with feathers auth
-        const sessionId = socket.handshake.headers.authorization.split(' ')[1]
+        const auth = socket.handshake?.headers?.authorization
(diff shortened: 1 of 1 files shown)


packages/server-core/src/sockets/sockets.ts
```diff
@@ -21,13 +21,23 @@ const handleSockets = (app: any) => {
...
+        const auth = socket.handshake?.headers?.authorization
+        if (!auth) {
+          return console.error(
+            'No Authorization header was provided in handshake'
+          )
+        }
+
+        const sessionId = auth.split(' ')[1]
+        if (!sessionId) {
+          return console.error('No session id provided for handshake')
```
michaelsharpe, line 33: Since this is a socket handshake request, we may want to send back the error and maybe with a status to the calling application.  Socket.io may have a way to error out the socket connection properly.  Then we can handle that on the client, or at least surface a useful error there.  It may be enough to just run `socket.disconnect('unauthorized')` or similar.

michaelsharpe (changes requested):

PR #542, opened 2023-03-26
Switch away from getCurrentSpell for socket events

michaelsharpe:
Should work now.  Brought up issue though that "getCurrentSpell" may not be a super reliable way to fetch the current spell. We may want to look into swapping it out in other places.  

The current spell is not injected on every run directly onto the whole context, rather than going through a helper function of the magick interface,


michaelsharpe:
consider this a draft.  have to fix some typing and remove all dependencies on getCurrentSpell in favor of the injected 'currentSpell' on the context.

PR #572, opened 2023-04-03
added collapse project window and comment item on context menu

FabriceIRANKUNDA:


FabriceIRANKUNDA:
To hide and show project window press control + b

PR #597, opened 2023-04-06
fixed infinite rerendering that was making dropdown to glitch

FabriceIRANKUNDA:


FabriceIRANKUNDA:
https://user-images.githubusercontent.com/58092199/230443374-ab8d2697-0357-458e-9459-285c3af52709.mov

This was mainly the problem for the dropdown to glitch
My inspection was that URI from useParams was not exactly equal to the activeTab.URI, when the same tab is opened which was causing the useEffect to keep, rerender that same spell endlessly.

I hope the trick I used will handle the issue, feel free to add any improvement you may come up with

michaelsharpe:
Looks great!

PR #618, opened 2023-04-11
Remove ignore auth but needs default token

lalalune:
This is a draft PR - do not merge

This removes the local development IGNORE_AUTH env variable and checks.

However, we need to instead supply a default token if the token is null and we are in a development / non production environment.

The reasoning here is that we want our local dev environment to handle auth the exact same as our cloud. Currently it doesn't, and a few of the windows in the cloud have auth issues. In order to avoid guess-and-check, we should make it the same.

- [x] Remove IGNORE_AUTH from everywhere
- [x] Add dummy token
- [ ] Test in cloud


coffeeorgreentea:
closed for for https://github.com/Oneirocom/Magick/pull/811#pullrequestreview-1418906017

PR #631, opened 2023-04-12
added project window to window menu options

FabriceIRANKUNDA:
You can show up project window via window menu option or using hotkey


FabriceIRANKUNDA:
https://user-images.githubusercontent.com/58092199/231552738-60e854db-274d-40de-aac5-b3fb375939de.mov

PR #700, opened 2023-04-20
Remove old build instructions.

fire:
The current build instructions are:

```
npm install --global nx
npm install --global knex
git clone https://github.com/Oneirocom/Magick magick -b development
cd magick
npm install
npm run dev
```

Remove the old build instructions.


fire:
Parzival> The main readme in project root should be the place where build instructions for getting started should be, and should be up to date.

PR #704, opened 2023-04-21
Enable posthog for better feedback from our users

coffeeorgreentea:
This adds posthog tracking to allow us the get better feedback from our users.
It will be enabled if the correct .env are set so our open source users need not to worry.
Envs are:
```
VITE_APP_POSTHOG_ENABLED=false
VITE_APP_POSTHOG_API_KEY=key
```

I have setup the .envs on dev/preview.
You can test it by using AIDE on this deployment and logging into the posthog account verifying it working.


coffeeorgreentea:
> Awesome. Can you please add
> 
>     * What this adds
> 
>     * How to test it
>       Will have @Knar33 review

Suer! I updated the original comment

PR #705, opened 2023-04-21
played with the IDE and made some clean up

FabriceIRANKUNDA:


lalalune:
@FabriceIRANKUNDA looking good, can you please write up
- What was changed
- How to test it
We have a new engineer who will be QAing PRs @Knar33 so I will pass this to him :)

Knar33 (approved):

Knar33:
> > Looks good to me, the import/export menu closes after clicking an option instead of staying open.
> 
> Is that behavior better than if it closes? Does it feel weird?

It feels weird when it stays open (old UX) especially when exporting, now you get a bit more feedback and you don't have to find somewhere to click to close the menu. I'm assuming people will normally import only one project at a time.

PR #708, opened 2023-04-24
Avoid error in FindTextEmbedding.ts

fire:
Error has logs.


fire:
Wanted to show @coffeeorgreentea. Archiving now.

PR #711, opened 2023-04-24
Remove chat options from the text generation node.

fire:
Was causing developers to get caught on a confusing error.

### DIFF
--- a/packages/core/shared/src/nodes/text/GenerateText.ts
+++ b/packages/core/shared/src/nodes/text/GenerateText.ts
@@ -54,7 +54,7 @@ export class GenerateText extends MagickComponent<Promise<WorkerReturn>> {
     // get completion providers for text and chat categories
     const completionProviders = pluginManager.getCompletionProviders('text', [
       'text',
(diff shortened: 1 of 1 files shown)


lalalune:
The most popular models, like 99% of what people use, is ChatGPT and GPT-4, which are the chat models, so we need to fix this :'D

coffeeorgreentea:
> The most popular models, like 99% of what people use, is ChatGPT and GPT-4, which are the chat models, so we need to fix this :'D

going to take a look into and try to fix rather then merge this.
@fire what exactly is the error?

fire:
I think it does not output. As far as I know chatgpt4 is not open to api usage. Not sure about gpt3.

lalalune:
> I think it does not output. As far as I know chatgpt4 is not open to api usage generally. Not sure about gpt3.

We all have GPT4, ChatGPT is open to all, it's 1/10th the cost of Davinci / GPT 3 so it's what almost all of our users use, and what we need to offer by default.

PR #712, opened 2023-04-24
Fix scrolling immediately to bottom of the console when expanding a node in the console

FabriceIRANKUNDA:
## What was fixed
When expanding a node in the console, it was scrolling immediately to bottom of the console


lalalune (approved):
This looks good to me, but needs some testing @Knar33

Knar33:
I just tested and it works, though the divs are all getting the same id "fabrizo" which is invalid html.

FabriceIRANKUNDA:
> I just tested and it works, though the divs are all getting the same id "fabrizo" which is invalid html.

Ooh! I added it for testing purposes, I'll remove it!

PR #717, opened 2023-04-24
Created a basic pull request template

Knar33:
## What Changed:

I added a basic pull request template to the root of the project 

## How to test:

I don't think we can test this, it has to be in the default branch in order to be used


fire:
See also https://github.com/Oneirocom/Magick/pull/720

Knar33:
> See also #720 which includes a form.

It seems like you can only use YAML for  issue forms.

PR #725, opened 2023-04-26
Using Elevenlabs API in avatar plugin

ultrabot2022:
## What Changed:

On the Avatar plugin, I have updated with Elevenlabs API to get audio.
https://api.elevenlabs.io/v1/

## How to test:

You can test by inputting text in the chatbox, can see the echo back of Eliza, her mouth movement and listen her voice.
Screenshot_6

Thank you for your time :)


plugins/avatar/client/src/components/Chat.jsx
```diff
@@ -21,6 +22,8 @@ const voices = {
 // Constants
 
 const messagesMaxCharacters = 20000
+const API_KEY = 'ce69df07b50e7179cbbfc5c2bef9d752';
```
lalalune, line 25: Can we please convert these to secrets and not put them in the public IDE (or in github) 🤣

PR #726, opened 2023-04-26
Fixed add/remove input sockets and reflect changes in the javascript code editor

FabriceIRANKUNDA:
## What Changed:

- [x] Fixed adding/removing input sockets
- [x] Reflect changes in the code editor by passing input sockets as arguments to the function

## How to test:
- Create code node
- Try Adding/Removing input sockets to it and ensure that changes are being reflected in the code editor as well accurately.
- You can refer to the video attached in the comment

## Additional information:

- For Python code node improvement is needed (I will talk to Moon about it)

https://user-images.githubusercontent.com/58092199/234518473-572d069c-8dcd-4d5d-9805-7f9fac6d30eb.mov

### DIFF
--- a/packages/core/shared/src/plugins/inspectorPlugin/Inspector.ts
+++ b/packages/core/shared/src/plugins/inspectorPlugin/Inspector.ts
@@ -4,7 +4,15 @@ import { v4 as uuidv4 } from 'uuid'
 
 import { MagickComponent } from '../../engine'
 import * as socketMap from '../../sockets'
-import { AsDataSocket, AsInputsAndOutputsData, DataSocketType, IRunContextEditor, MagickNode, PubSubData, WorkerData } from '../../types'
+import {
+  AsDataSocket,
+  AsInputsAndOutputsData,
+  DataSocketType,
+  IRunContextEditor,
+  MagickNode,
+  PubSubData,
+  WorkerData,
+} from '../../types'
 import { DataControl } from './DataControl'
 
 type InspectorConstructor = {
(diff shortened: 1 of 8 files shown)


Knar33:
I'm testing using this setup:
image

Adding/removing nodes does update the function parameters as described. 

A couple bugs I found in the js code node:
- I'm having trouble getting code to save on a new js code node when it has no input sockets. If I drop in an empty code node, change things in the text editor, save text editor, save spell, then refresh, the code reverts back to the default code. One time when I dropped in a fresh js code node, every time I hit save it would revert the changes I made in the text editor (this only happened to me once and I can't replicate it). Changes start persisting when I add an input socket, save the spell, and refresh. If I remove the input socket saving in the editor works, it's just the initial state that seems broken.
- When adding new input sockets to a new code node it will erase all of the code I've written in the code editor. I think this is related to the above, because adding an input socket + saving + refreshing seems to make the bug go away. 
- If I click on my "text variable" node, then click back to the js code node, it seems to replace the text inside the javascript text editor with the text from the text variable. It doesn't actually override the code but if you hit "save" it will break the text editor for the js node (there's no text box, refreshing doesn't bring it back)
image

FabriceIRANKUNDA:
@Knar33 when switching between "text variable"  node and the code node, It is not supposed to override the code instead it shows the text from the text variable node.

Knar33:
@FabriceIRANKUNDA Looks like you fixed all 3 of the issues I mentioned, except now the Text Variable node never shows the text editor. I tried it with a fresh spell and I still can't enter any text into the node, the text box never shows up.

Knar33 (approved):
The Javascript code node is working smoothly now. I edited the PR title to note that this is only effecting the Javascript code editor.

PR #727, opened 2023-04-26
Add keyboard commands alt+[key] and ctrl+[key] to feel more like Unreal Engine

coffeeorgreentea:
## What Changed:

Keyboard commands for options in dropdown menus are working and added.
They are listed as alt+[key] but takes a greedy approach and accepts ctrl+[key] where available.

## How to test:

Try the shortcuts listed in the file menu options.


packages/editor/src/plugins/contextMenu/react-menu/Menu.tsx
```diff
@@ -36,14 +36,18 @@ export default function ContextMenu({
...
-      searchbarRef.current.focus()
+    if (visible) {
+      const searchbar = document?.querySelector('.context-menu-search-bar')
+      if (searchbar) searchbar.focus()
+    } else {
+      setSearch('')
     }
   }, [visible])
-
+
```
fire, line 50: Whitespace oh no!

Knar33:
All of the hotkeys work except for ctrl+N. I tested in Firefox but I believe the same issue will be in Chrome. They don't allow scripts to override the ctrl+N hotkey.

It might be worth leaving in ctrl+shift+z as a hidden hotkey for Redo, there are a few editors like Blender that use that key combination.

coffeeorgreentea:
> All of the hotkeys work except for ctrl+N. I tested in Firefox but I believe the same issue will be in Chrome. They don't allow scripts to override the ctrl+N hotkey.
> 
> It might be worth leaving in ctrl+shift+z as a hidden hotkey for Redo, there are a few editors like Blender that use that key combination.

Updated! ctrl/alt/option+shift+z is in there now good idea. It also will display correct commands for linux now. You can test the other OS display by changing your user agent in your browsers insepctor. Also cleaned up a file that also registered some hotkeys that were duping.

Knar33 (approved):
Looks good! All the hotkeys work as well as the alternate hidden versions using ctrl modifier, ctrl+shift+z does redo, and when I change my user agent to something without windows/linux in the name it changes the hotkeys modifiers from alt to option.

PR #736, opened 2023-04-27
Babyagi node added

sshivaditya2019:
## What Changed:

The new BabyAGI node has a dependency on the Agent SpellManager, which is responsible for executing the TaskCreation, TaskReprioritization, and TaskExecution spells required by BabyAGI.

## How to test:

To test the process, include the spells for Task Creation, Task Reprioritization, and Task Execution. The remaining steps can be tested as usual.

## Additional information:

Each iteration of BabyAGI involves three API calls to OpenAI.

This is a MVP. This will not on work on cloud yet.


coffeeorgreentea:
@sshivaditya2019 i butchered a conflict resolve so moving this PR to here where fire undid my commit https://github.com/Oneirocom/Magick/pull/737

sshivaditya2019:
Task Creation, Task Reprioritization, and Task Execution spells, have to be imported before the BabyAGI node is called.

Knar33:
@sshivaditya2019 I have the spells loaded and a discord enabled agent running the BabyAGI sopell, but I'm not sure if I understand how it's supposed to work. It takes a minute and cycles through a bunch of different versions of its task list, and ends up sending just one message in discord right before it stops. Is that behavior expected? Is there a way to advance to the next task?

sshivaditya2019:
Baby AGI, essentially should run till infinity, where it keeps on exploring new ideas and tasks. Here, I am restricting it to run 2 loops, after which it will output in the same discord channel, the task it executed the last. This is the intended behavior,  it can be made to explore further by increasing the number of iterations.

lalalune:
I'm going to merge this to keep integration simple. We can add to it. For now it does a basic version of babyagi.

PR #737, opened 2023-04-27
Add baby agi node.

fire:
Fixes: https://github.com/Oneirocom/Magick/issues/738

## What Changed:

Amended version of https://github.com/Oneirocom/Magick/pull/736

## How to test:

Use the same tests as https://github.com/Oneirocom/Magick/pull/736

## Additional information:


michaelsharpe:
@sshivaditya2019 We should hard code the JSON for these into the code base.  Rather than getting the spells in the worker from the DB, we can import json files for the spells.  This will let it work for anyone in theory.

PR #759, opened 2023-04-30
Fix: Babyagi Node

### LINKED ISSUE #738: BabyAGI like demo
See https://github.com/Oneirocom/Magick/pull/736 and https://github.com/Oneirocom/Magick/pull/737

This is a MVP. This will not on work on cloud yet.

sshivaditya2019:
## What Changed:

The number of iteration output is taken from the Inspector panel.

## How to test:

Using the same spell shared before

## Additional information:

Fixes #738


Knar33:
It seems like it's ignoring the iterations. I ran it with the iterations set to 1, and then 3, and compared how many times the text "You are a task creation AI" showed up in the terminal log and it was same number of times (6). When I changed the iterations I had the bot turned off, changed iterations and saved spell, then turned the bot back on. 

I also compared the number of requests sent to OpenAI API
1 iteration:
image

3 iterations:
image

I would expect there to be more requests for 3 iterations.

lalalune:
Merged into `eliza` so closing

PR #776, opened 2023-05-03
Adds overflow property to dropdown text

benbot:
## What Changed:

Added `overflow: clip` to dropdowns. See attached screenshots

Before
before_pr

After
after_pr


fire (approved):
We talked about how the box can grow, but clipping is ok.

PR #778, opened 2023-05-03
fix gpt forgetting directive

### LINKED ISSUE #735: Starter discord bot acts as if it's ignoring the system directive
### Magick version

development a6c8eb1

### System information

Ubuntu 22.04.1

### Issue description

When using the starter discord bot spell, the first few messages follow the system directive, but once you have a bunch of events to send it seems like the system directive is completely ignored. This may be related to #732 and #733 because too many events are being sent to chatgpt, but there's a chance that sending the system directive as the last message in the messages array will give it a higher weight (according to a quick google search about the issue).

### Steps to reproduce

- Create a starter discord bot
- Give it a system directive
- Send a few messages about things unrelated to the system directive
- Ask the discord bot "Who are you?"

### Minimal reproduction project

_No response_

benbot:
Should resolve #735 

## What Changed:

The system message is now always the last message in a conversation list sent to openai. 

This seems to prevent gpt from forgetting our directive and doesn't seem to affect the output at all (other than making it more strictly following the directive)

## How to test:

In the current development branch

Create a new discord bot spell.
Have a small conversation with it about AI.
I asked it to add some numbers then asked it how to explain AI to my boss.

Observe that after a few messages it seems to forget that its supposed to be a towel.

In my branch

follow the same steps and observe that it never forgets that its a towel.

### DIFF
--- a/packages/core/shared/src/nodes/text/TextVariable.ts
+++ b/packages/core/shared/src/nodes/text/TextVariable.ts
@@ -1,8 +1,8 @@
-// DOCUMENTED 
+// DOCUMENTED
 /**
  * Module represents a Rete flow based on Google code standards.
  * @module
- */ 
+ */
 
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import Rete from 'rete'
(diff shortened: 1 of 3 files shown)


packages/core/shared/src/nodes/text/GenerateText.ts
```diff
@@ -179,7 +179,7 @@ export class GenerateText extends MagickComponent<Promise<WorkerReturn>> {
       throw new Error('ERROR: Completion handler undefined')
     }
 
-    const { success, result, error } = await completionHandler({
```
lalalune, line 182: The completion handlers this is calling are async, so I think we need this await

packages/core/shared/src/nodes/text/TextVariable.ts
```diff
@@ -70,16 +70,16 @@ export class TextVariable extends MagickComponent<InputReturn> {
...
+   * @param {MagickWorkerInputs} _inputs - The inputs of the node. In this case, this object is
    * not used.
    * @param {MagickWorkerOutputs} outputs - The possible outputs of the node.
    * @param {Object} context - The data passed to the worker and the module.
-   * @returns {InputReturn} - The outputs of the node. In this case, an object with the output 
+   * @returns {InputReturn} - The outputs of the node. In this case, an object with the output
    * string.
    */
-  worker(node: WorkerData, inputs: MagickWorkerInputs, 
+  worker(node: WorkerData, _inputs: MagickWorkerInputs,
```
lalalune, line 82: Is this a functional change or just formatting?

lalalune (commented):
The change is fine but the other non-necessary changes might cause bugs, see comments.

packages/core/shared/src/nodes/text/TextVariable.ts
benbot: It stops the linter from complaining, just a formatting change as `input` is unused.

packages/core/shared/src/nodes/text/GenerateText.ts
benbot: Oh i see.... typescript was saying that the `await` has no effect there. Oh i see, the handler definition doesn't say that it returns a promise. let me get that change in too.

benbot: okay changed the type definition to include promises and not promise return values

lalalune (approved):

Issue #399, opened 2023-03-13
Subspells working in agents


lalalune:
Subspells are not currently handled by agents but should be

- [ ] Expose public variables of subspells to graph
- [ ] Standardize event input/output
- [ ] Cache spells on agent
- [ ] Updated agent checks for spells needing to be reloaded

PR #781, opened 2023-05-04
Agent architecture update and subspell fixes

michaelsharpe:
## What Changed:

- Added redis support for syncing multiple instances of the server, and sending messages from agent server to main server
- redis should only be activated for now by presence of a redis-url.  Not having it shouldn't break it.
- Switched spell runner architecture.  SpellManager is now passed into spellRunner, and all spellRunners need to be made by a spellManager
- added event logging to agent service and agent so we can stream events out
- fixed Spell component to use the new injected SpellManager to run spells
- fixed subSpell running in both IDE preview and in agent

## How to test:

- ensure normal spells run as normal in preview and in an Agent
- test subSpells.  Make a spell, use it in the Spell component of the root spell.  Run in a live agent via discord, and in the IDE preview
- ensure that everything runs even without presence of the REDIS_URL in the environment
- 

## Additional information:

- you will need to npm install some new dependencies
- this introduces some deep changes to how spells are run in general, so it is important to ensure that all things involving spell running are operating as normal


michaelsharpe:
Partly fixes #399 Does not add fixes for agent checks for spells needing to be reloaded or for Cache spells on agent

lalalune:
We should pull this in now since it adds the redis stuff, and merge things on top of it

PR #784, opened 2023-05-04
change inspector ports so that server and agent don't clash anymore

benbot:
## What Changed:

both agent and server were running their inspectors on port 7778, now server is on 7778 and agent is on 7779

## How to test:

check that both inspector servers are running on their correct parts.


fire:
I endorse this change. Going to run a `npm run dev` and report back.

Issue #744, opened 2023-04-28
Starter discord bot spell not showing stored events in the event UI


Knar33:
### Magick version

ccee4d4

### System information

Ubuntu 22.04.1, firefox

### Issue description

With the starter discord bot spell, when I chat with it in Discord I do not see any events show up in the Events page. It is storing and recalling the events though, I can see them in the API requests, they just don't show up in the GUI. When I use the playtest in the editor it properly stores my events. 

### Steps to reproduce

- Create a discord bot starter spell
- Chat with it in Discord
- Check the Events page for your events

PR #791, opened 2023-05-05
Misc bug fixes (loading screen, project window, events, discord)

FabriceIRANKUNDA:
## What Changed:
- Added a loading screen when fetching events and documents
- Fixed project window avoiding interaction with controls in the inspector
- Events are using maxcount and conversation type when filtering
- Now Discord is sending messages based on the role of either assistant or user

## How to test:
- use hotkey ⌃ + B to show up project window and try interacting with any control in the inspector
- create event recall and change type and maxcount properties to something else rather than conversation and inspect the type and size of events that are returned

## Additional information:

For issue #744 events are being shown in the event UI when you're not using local database sqlite instead use the postgres


Knar33:
- project window now longer messes with the inspector
- Event recall is bringing in the proper number of events
- Event recall is grabbing the correct Type of events
- The role of user is being sent to openAI correctly
- I can briefly see a "please wait" grey screen on documents/events pages (I had to set cpu throttling to 6x to get it to show up).

Only issue with this is that Events isn't grabbing the latest events. It's grabbing the oldest events, so it always returns the same n events.

Knar33 (approved):
I committed a fix the the one issue I found. Looks good now!

PR #792, opened 2023-05-05
updates docker script to use the new docker compose command

benbot:
## What Changed:

docker-compose is deprecated and hits EOL in june this year, so this change keeps us up to date
https://docs.docker.com/compose/compose-v2/

### DIFF
--- a/scripts/check-docker.js
+++ b/scripts/check-docker.js
@@ -1,2 +1,23 @@
-// check if docker is running
-const dockerStart = require('child_process').execSync('docker-compose up -d');
+const tryV2 = () => {
+  try {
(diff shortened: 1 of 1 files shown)


michaelsharpe:
My one concern with this is that to use it with docker desktop a user has to go into docker desktop and turn on a 'use docker compose v2' settings. My hope had been a less technical user would just have to download docker desktop and that would be enough. This will require an added step of complexity for a user to get started from the repo.

michaelsharpe:
@benbot perfect! Thanks to much. Future proofs us. And yeah I imagine when they sunset v1 they will make v2 the desktop standard.

PR #809, opened 2023-05-07
Fix windows build

coffeeorgreentea:
## What Changed:

Building on Windows was not working because the dev script was only setting NX_NON_NATIVE_HASHER=true for mac/linux. I fixed this by using cross-env. Also added a nx-win32 as a optional dependency to fix another error.

## How to test:

Build and run Magick.

## Additional information:

n/a


Knar33:
It still builds and runs for me on Linux. I don't have a Windows machine I can test this on.

PR #811, opened 2023-05-08
Remove ignore auth but needs default token 2

coffeeorgreentea:
This is a draft PR - do not merge
Related: https://github.com/Oneirocom/Magick/pull/618
I ended up making a new branch for this.

This removes the local development IGNORE_AUTH env variable and checks.

However, we need to instead supply a default token if the token is null and we are in a development / non production environment.

The reasoning here is that we want our local dev environment to handle auth the exact same as our cloud. Currently it doesn't, and a few of the windows in the cloud have auth issues. In order to avoid guess-and-check, we should make it the same.

- [x] Remove IGNORE_AUTH from everywhere
- [x] Add dummy token
- [x] Tested locally

To test we need to make sure each page and all requests are working both in the cloud with production set to true and in local when production is set to false.

Also to note i signed the dummy token for a year and its signed with the default secret: 'secret'.


Knar33:
I was able to access every page when running Magick locally. 

When I run Cloud (with Magick .env PRODUCTION=true  and VITE_APP_PRODUCTION=true) I get this error, even after signing out and signing back in:
NotAuthenticated: decryption operation failed
    at CloudJwtService.verifyAccessToken (/home/knar/Documents/Github/Magick/dist/apps/server/webpack:/packages/core/server/src/auth/cloudAuthentication.ts:76:13)
    at JWTStrategy.authenticate (/home/knar/Documents/Github/Magick/node_modules/@feathersjs/authentication/src/jwt.ts:140:21)
    at CloudJwtService.authenticate (/home/knar/Documents/Github/Magick/dist/apps/server/webpack:/packages/core/server/src/auth/cloudAuthentication.ts:93:20)
    at /home/knar/Documents/Github/Magick/node_modules/@feathersjs/authentication/src/hooks/authenticate.ts:56:26
    at SpellService.logError (/home/knar/Documents/Github/Magick/dist/apps/server/webpack:/packages/core/server/src/hooks/index.ts:12:5)
    at /home/knar/Documents/Github/Magick/node_modules/@feathersjs/koa/src/rest.ts:36:21
    at /home/knar/Documents/Github/Magick/node_modules/@feathersjs/koa/src/handlers.ts:6:5
    at cors (/home/knar/Documents/Github/Magick/node_modules/@feathersjs/koa/node_modules/@koa/cors/index.js:108:16)
Data: { code: 'ERR_JWE_DECRYPTION_FAILED', name: 'JWEDecryptionFailed' }

coffeeorgreentea:
@Knar33 Did you force a rebuild? I have to for envs that are used through config.ts

Knar33 (approved):
My issue was not having the secret set to the same value. I just tested with the correct secret value and everything worked as described.

PR #812, opened 2023-05-08
changed enable agent button and fixed a bug with agent secrets

### LINKED ISSUE #687: [BUG] Two agents with same OpenAI API key respond differently
If I make two agents from the same root spell and assign them the same OpenAI key, it seems one is returning a 401 and the other one works.  I need to try and track down if this is replicable, but I recall another user bringing this one up too.

FabriceIRANKUNDA:
## What Changed:

- Changed agent enable button and remove unnecessary action to hit the update button after hitting enable button
- Fixed agent secrets when a new agent is created to fix #687
- Fixed updating public variables at agent page

## How to test:

- Try enabling/disabling the agent
- create a new agent and configure the same open ai as the other agents are using
## Additional information:

### DIFF
--- a/apps/docs/docs/api/index.md
+++ b/apps/docs/docs/api/index.md
@@ -60,7 +60,7 @@ The interface for an agent object that's based on the `agentSchema`.
 
 #### Defined in
 
-[packages/core/shared/src/schemas.ts:70](https://github.com/Oneirocom/Magick/blob/11f4cdf5/packages/core/shared/src/schemas.ts#L70)
+[packages/core/shared/src/schemas.ts:75](https://github.com/Oneirocom/Magick/blob/966ea52f/packages/core/shared/src/schemas.ts#L75)
 
 ___
 
@@ -72,7 +72,7 @@ The type for an agent object that's based on the `agentSchema`.
 
 #### Defined in
 
-[packages/core/shared/src/schemas.ts:68](https://github.com/Oneirocom/Magick/blob/11f4cdf5/packages/core/shared/src/schemas.ts#L68)
+[packages/core/shared/src/schemas.ts:73](https://github.com/Oneirocom/Magick/blob/966ea52f/packages/core/shared/src/schemas.ts#L73)
 
 ___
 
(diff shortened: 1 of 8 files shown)


Knar33:
The enable button automatically turns on the agent without hitting update.

What changed with public variables exactly? I wasn't sure what to test here.

I tried running my discord bot but I'm getting an error when saving events, may not be a new error though. I can see that both of my bots (using the same root spell and same openai key) get a successful text completion from OpenAI though.

Knar33:
> The enable button automatically turns on the agent without hitting update.
> 
> What changed with public variables exactly? I wasn't sure what to test here.
> 
> I tried running my discord bot but I'm getting an error when saving events, may not be a new error though. I can see that both of my bots (using the same root spell and same openai key) get a successful text completion from OpenAI though.

The error I'm getting is not new to this pull request, it's happening in Development also. 

Is the bug you fixed with multiple agents supposed to happen when you have two discord bots running at the same time? I was testing it with only one active at a time. I didn't get the 401 error in Development either so I might not be testing this one correctly.

FabriceIRANKUNDA:
> > The enable button automatically turns on the agent without hitting update.
> > What changed with public variables exactly? I wasn't sure what to test here.
> > I tried running my discord bot but I'm getting an error when saving events, may not be a new error though. I can see that both of my bots (using the same root spell and same openai key) get a successful text completion from OpenAI though.
> 
> The error I'm getting is not new to this pull request, it's happening in Development also.
> 
> Is the bug you fixed with multiple agents supposed to happen when you have two discord bots running at the same time? I was testing it with only one active at a time. I didn't get the 401 error in Development either so I might not be testing this one correctly.

Basically, having two discord bots with the same OpenAI secret is not causing any problem, the issue was when creating a new discord bot the open ai key input field was showing value from a previously selected agent and that was reflecting as if open ai key has already been configured while it has not been configured which was causing a 401 error.

Knar33 (approved):
Looks good. Creating a new agent has a blank openAI secret field, enabling agents works without the extra update click, and the textbox for changing public variables from the agent screen works properly now.

PR #825, opened 2023-05-10
Integrate WebXR and LookingGlass with VRM support and improve UX

benbot:
## What Changed:

Update scene for watching, drag/drop and init XR option.

## How to test:

Ultrabot plz fill out

Replaces #824


Knar33:
- The microphone seems to work, but I was a little confused with the icons. When there is a slash through the microphone that means it's unmuted? seems backwards.
- When I drag and dropped VRM files they were facing the wrong direction
- The VRM does look at the cursor
- I'm not sure how to test the webxr stuff, do I need a vr device?

fire (approved):

fire:
@ultrabot2022 

Need to review these issues, but also need to schedule them given our milestone 1 priorities.

-     The microphone seems to work, but I was a little confused with the icons. When there is a slash through the microphone that means it's unmuted? seems backwards.
-     When I drag and dropped VRM files they were facing the wrong direction
-     The VRM does look at the cursor
-     I'm not sure how to test the webxr stuff, do I need a vr device?

ultrabot2022:
> The microphone seems to work, but I was a little confused with the icons. When there is a slash through the microphone that means it's unmuted? seems backwards.
I will complete this after getting response from the spell.

> When I drag and dropped VRM files they were facing the wrong direction
It is the vrm file problem. Also I will implement option for any vrm files to face same direction on the next milestone.

> I'm not sure how to test the webxr stuff, do I need a vr device?
I am using WebXR API Emulator extension of chrome browser :)

PR #831, opened 2023-05-12
implementing event UI

FabriceIRANKUNDA:
## What Changed:

Added new event UI

## How to test:

Visit event window from the drawer

## Additional information:


lalalune:
@FabriceIRANKUNDA I linted this so it would pass, even though it's draft.

PR #841, opened 2023-05-15
Fixed search and resized loading screen

FabriceIRANKUNDA:
## What Changed:

- Fixed search 
- Enable sort on column
- resized loading icon
## How to test:

- Search event
- click on column to sort asc or desc on that column
## Additional information:


lalalune:
Merged all latest. This PR is currently broken when running the client IDE standalone locally, because the dummy token is out of date. We need to generate a new dummy token on startup to prevent expiry, or we need a dummy token that doesn't expire.

lalalune:
Screenshot 2023-05-18 at 1 16 01 PM
Screenshot 2023-05-18 at 1 15 55 PM
Screenshot 2023-05-18 at 1 15 52 PM

PR #864, opened 2023-05-18
Fix authentication error on API and update API to new spell manager and run handler

lalalune:
## What Changed:
This PR fixes the API connector.

NOTE: In progress, I need to make sure POST works as well

To get GET

```
async function test() {
    const output = await fetch('http://localhost:3030/api/e552337d-3a88-4ac9-af2d-fec929bec4e4?apiKey=88bac817bd7a004d4a0f1a26b0fcf0a7&content=Hello+World')
    const json = await output.json()
    console.log(json)
}

test();
```

### DIFF
--- a/packages/core/server/src/app.ts
+++ b/packages/core/server/src/app.ts
@@ -134,6 +134,11 @@ app.hooks({
     all: [
       logError,
       async (context: HookContext, next) => {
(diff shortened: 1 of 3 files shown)


Knar33 (approved):
I created a test spell with an input for each of the four REST methods and all four worked.

PR #865, opened 2023-05-18
Increase feathers limits to 1000

coffeeorgreentea:
## What Changed:

Spells, documents, requests and events fetching has been increased to a limit of 1000.

## How to test:

Make sure all four of those still work, and try to see if the limit goes beyond 100 now.

## Additional information:

Pagination soon to follow in another issue.


Knar33 (approved):
Documents, events, and requests go past page 10 now. The Open Spell menu and the project window shows more than 100 spells and exporting my project exports more than 100 spells.

PR #906, opened 2023-05-21
Fix project imports

coffeeorgreentea:
## What Changed:

I accidentally committed something that broke project imports in spell.schema.ts

## How to test:

Import Eliza!

## Additional information:

Sorry about that!


Knar33 (approved):
Looks good, I can now import Eliza's spells!

PR #924, opened 2023-05-24
Delete electron.yml

fire:
## What Changed:

Solves package pinning security issues and the feature was never completed.

## How to test:

Check that nothing is broken.


lalalune:
The electron feature was completed and works, the part that isn't completed is having it build in CI. It's autobuilt by https://capacitorjs.com/

fire:
Thanks. I'll resolve the pinning in a different pr.

PR #934, opened 2023-05-24
Update electron.yml

fire:
## What Changed:

Pin electron actions using https://app.stepsecurity.io/secureworkflow/Oneirocom/Magick/electron.yml/development?enable=pin

## How to test:

Check if all things build.

### DIFF
--- a/.github/workflows/electron.yml
+++ b/.github/workflows/electron.yml
@@ -15,9 +15,9 @@ jobs:
         node-version: [19.x]
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


fire:
I believe they are the hashes of the github actions themselves like the entire package of https://github.com/actions/upload-artifact

The concern from OSSF is that this code can change and introduce unwanted behaviour automatically.

lalalune:
We should merge this. The electron build actions need some work anyways, at least this makes sure it's protected from supply chain attack

lalalune (approved):

PR #939, opened 2023-05-25
make logging enabling/disabling clearer

coffeeorgreentea:
## What Changed:

It wasn't very clear to local users how to disable logging, and its reliance on production made testing production reliant things harder.
This adds a flag specifically for logging.
## How to test:

Set the LOGGING_ENABLED .env to true or false depending on what you are testing.

## Additional information:

N/A


benbot (changes requested):
We should never disable logging, i don't think.

Pino has the concept of log levels, we should just have an env flag that sets those.

people who want to ignore all but the most serious logs can set the log level to fatal

PR #945, opened 2023-05-26
added max count on event recall when using embeddings

FabriceIRANKUNDA:
## What Changed:

- Added considering max count on event recall when using embeddings

## How to test:

## Additional information:


Knar33:
It is taking into consideration the max count, but it's not grabbing the top embedding.

image
 
I have this test rigged up to create 2 embeddings ("test" and "banana pudding") then I do the event recall using the exact embedding from "test", but instead it returns "banana pudding"

Knar33 (approved):

PR #946, opened 2023-05-26
Cloud agents

benbot:
POC for cloud agents

first crack at cloud agents.

I may look into streamlining the agent -> agentManager relationship too.

### Testing steps:

I made changes to the database, so be sure to run `npm run migrate`

instead of running `npm run dev` run `npm run cloud`. This will NOT start the existing agent server and will instead start the new CloudAgentManager and CloudAgent worker

enable an agent (this doesn't work with already enabled agents yet)
observe that the agent works.

disable that same agent
observe that the agent stops working


fire:
Can you add a how to test section and also a rough design

Knar33 (approved):

Issue #896, opened 2023-05-19
Event Delete node passes on its trigger before the event is fully deleted


Knar33:
### Magick version

aef2d50

### System information

Ubuntu 22.04.1, firefox

### Issue description

When you use the Event Delete node, it does delete the event but it passes on its trigger before the event is fully deleted. You can retrieve the event immediately afterwards and it will come through.

### Steps to reproduce

- Create the graph below
- Make sure that you have no events in your eventstore with the event type of "eventdelete"
- input any text into the playtest window and hit send
- See that the agent outputs a JSON string of the event that should not exist anymore

![image](https://github.com/Oneirocom/Magick/assets/4692561/38ee5fcd-7c9e-4b02-963e-002400990798)

To verify that the events are being deleted properly, string up the Create text embedding node between the event delete and event recall, and run the same test. You will get an empty array back. This what you should get back in the above test (the embedding node adds a bit of time delay to allow the event to fully delete)

![image](https://github.com/Oneirocom/Magick/assets/4692561/198ce8c0-840f-4586-86a1-f39202402c31)

### Minimal reproduction project

```JSON
{
    "id": "52799a0d-b4ac-40ce-90a1-c8b29224d99b",
    "name": "event delete breaking",
    "projectId": "bb1b3d24-84e0-424e-b4f1-57603f307a89",
    "hash": "eb753bd87886661cbb08fa48f0a7201d",
    "createdAt": "1684531721051",
    "updatedAt": "1684532142790",
    "graph": {
        "id": "demo@0.1.0",
        "nodes": {
            "232": {
                "id": 232,
                "data": {
                    "name": "Input - Default",
                    "text": {
                        "id": "2eb64348-adee-4118-bcca-19697ca3a16a",
                        "inputs": {
                            "Input - Default": {
                                "type": "playtest",
                                "client": "playtest",
                                "sender": "playtestSender",
                                "agentId": "preview",
                                "channel": "playtest",
                                "content": "testing",
                                "entities": [
                                    "playtestSender",
                                    "Agent"
                                ],
                                "observer": "Agent",
                                "projectId": "bb1b3d24-84e0-424e-b4f1-57603f307a89",
                                "channelType": "playtest"
                            }
                        },
                        "projectId": "bb1b3d24-84e0-424e-b4f1-57603f307a89",
                        "spellName": "Starter",
                        "publicVariables": "[]"
                    },
                    "isInput": true,
                    "outputs": [],
                    "success": false,
[...]

PR #950, opened 2023-05-27
Fix event delete node passes on trigger before event is deleted #896

lalalune:
Fixes https://github.com/Oneirocom/Magick/issues/896


Knar33:
I keep getting this error when I call the event delete node. It seems like the event delete works but it throws this error before passing on the trigger:
image

I tried casting the event to an object before passing it in and that didn't work.

Knar33:
I'm still getting an error on this node. Event recall is definitely getting an event array with one event, I'm taking the 0 element from the array and passing it into delete event, and this happens.
image

PR #952, opened 2023-05-28
Fix Twitter, Fix Plugins refresh and caching, move Plugins folder into Packages folder

lalalune:
This PR fixes Twitter and several paths and improves developer experience.

Because nx is only set up for "apps" and "packages" folder, moved the "plugins" folder into it.

Fixed circular dependency.

Fixed Twitter connector for feed input.


Knar33:
I got this error when doing npm run
image

lalalune:
@Knar33 should be fixed now

Knar33:
I was able to get the DM input to work, and I had it wired up to respond but I can't verify it did.

I had a feed input in the same spell and haven't seen it trigger yet, even though I tweeted @gpt4bot (you're not following me so maybe it didn't show up?).

I restarted my bot and got this error which is preventing further testing:
image

packages/core/client/src/components/Drawer/ProjectWindow/index.tsx
```diff
@@ -105,7 +104,7 @@ const ProjectWindow = ({ openDrawer }) => {
     const element = document.createElement('a')
 
     const exportData = data
-    exportData.agents.forEach((agent: Agent) => {
+    exportData.agents.forEach((agent: any) => {
```
benbot, line 107: Why even use typescript at this point :'(

packages/core/shared/src/nodes/io/Respond.ts
```diff
@@ -80,6 +80,7 @@ export class Respond extends MagickComponent<void> {
     _outputs: MagickWorkerOutputs,
     context: ModuleContext
   ): Promise<{ output: string }> {
+    console.log('calling respond')
```
benbot, line 83: same note in this file about using the pino logger instead of the console log

packages/core/shared/src/spellManager/SpellManager.ts
```diff
@@ -11,7 +10,7 @@ type SpellManagerArgs = {
   cache?: boolean
   app: Application
   watchSpells?: boolean
-  agent?: Agent
+  agent?: any
```
benbot, line 13: Is there any way we can maintain our types? Can we make this a `Agent | something` type?

packages/editor/src/contexts/MagickInterfaceProvider.tsx
```diff
@@ -209,7 +209,7 @@ const MagickInterfaceProvider: React.FC<{
     const response = await _runSpell({ inputs, spellId, projectId })
 
     if ('error' in response) {
-      throw new Error(`Error running spell ${spellId}`)
+      throw new Error(`Error running spell from runSpell ${spellId}`)
```
benbot, line 212: can we have a `logger.error` line here as well?

packages/plugins/bluesky/server/src/index.ts
```diff
@@ -46,7 +46,6 @@ function getAgentMethods() {
 
 async function handleResponse({ output, agent, event }) {
   console.log('********* SENT MESSAGE TO BLUESKY', agent.id, output, event)
```
benbot, line 48: these lines weren't added in this PR, but eventually we should move these over to being pino logs too

benbot (changes requested):
The actual code changes look good to me. Just some housekeeping changes, really.

Can we try replacing the `console.log` with pino logs, so we can track them more easily?

And is there any way to keep the Agent type instead of changing it to any? it kind of defeats the purpose of typescript.

In the future, we might want to separate out large refactor PR from bugfix PRs, so we don't have 766 files to look over :sweat: Luckily most of them were just files being moved around and configs being changed.

Knar33:
> I guess also kind of a nitpick (nothing needs to be changed), but why was the fix-delete-agent branch merged into this ones instead of just being its own PR?

It was its own PR, which got merged already.

Knar33 (approved):
The twitter bot works. Not sure what else to test.

packages/core/server/src/services/events/events.ts
```diff
@@ -53,7 +53,7 @@ export const event = (app: Application) => {
         (context: HookContext) => {
           const { getEmbedding } = context.params.query
           if (getEmbedding) {
-            context.params.query.$limit = 1
+            // context.params.query.$limit = 1
```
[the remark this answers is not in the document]
lalalune: We should definitely not be hard-limiting document queries, this could be a separate PR but this is a bug

packages/core/shared/src/spellManager/SpellManager.ts
lalalune: The problem is dependency order, can't use agent here because of circ dep

benbot: > The problem is dependency order, can't use agent here because of circ dep

can we break out where the agent type is defined?

benbot (approved):
LGTM

packages/core/shared/src/spellManager/SpellManager.ts
lalalune: The correct solution here needs to be applied more globally, which is that all types need to be in the packages/@types folder, and we have a type for Agent that all of these are inheriting from

Knar33 (approved):

Knar33 (approved):

benbot (dismissed):

packages/core/shared/src/nodes/io/Respond.ts
```diff
@@ -96,21 +96,28 @@ export class Respond extends MagickComponent<void> {
...
         return type.name.includes(
           (event.connector as any).replace('Input - ', '')
         )
       })
+      console.log('type', type)
 
       const responseOutputType = type?.defaultResponseOutput
       const out = agent.outputTypes.find(t => t.name === responseOutputType)
 
+      console.log('responseOutputType', responseOutputType)
```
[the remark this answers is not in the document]
benbot: I have a codeshift in the works to remove stray console logs coming, so we can leave this for nwo

benbot (approved):

PR #953, opened 2023-05-29
Allow text-editor to save code

alexjorgef:
## What Changed:

On the frontend Monaco editor, the problem is that the code block only updates using _Python_ and _Javascript_ nodes (from **Code** category). This change allows the code block to update for other categories.

### DIFF
--- a/packages/editor/src/windows/TextEditorWindow/index.tsx
+++ b/packages/editor/src/windows/TextEditorWindow/index.tsx
@@ -49,7 +49,7 @@ const TextEditor = props => {
 
   useEffect(() => {
     if (!inspectorData?.data.inputs || !inspectorData?.data.inputs.length) {
-      if (inspectorData?.category !== 'Code') {
+      if (inspectorData?.category === 'Text') {
         setCode(inspectorData?.data?.fewshot)
         return
       }


packages/editor/src/windows/TextEditorWindow/index.tsx
```diff
@@ -49,7 +49,7 @@ const TextEditor = props => {
 
   useEffect(() => {
     if (!inspectorData?.data.inputs || !inspectorData?.data.inputs.length) {
-      if (inspectorData?.category !== 'Code') {
+      if (inspectorData?.category !== 'Code' && inspectorData?.category !== 'Ethereum') {
```
benbot, line 52: This change looks fine to me, but seeing these hardcoded strings makes me uneasy.

Is there any way we can extract these to be named constants somewhere?

alexjorgef:
@benbot Rebased and add **Database** category

benbot (changes requested):
Could that array remain as a type?

alexjorgef:
@benbot I revert the type definition. And consider the fewshot code just the `Text` category nodes. This will permit other categories to use the code editor

benbot (approved):

PR #954, opened 2023-05-29
fixed various bugs

FabriceIRANKUNDA:
## What Changed:

- fixed logical operator wrong operation type
- Is Null or Undefined node is fine
- Fixed Get Values from the Array node label
- Fixed number variable output type
- Fixed cast node
- Fixed Text rule matcher Minimum/Maximum
- Fixed merge objects throwing an error when don't pass an optional object
- Fixed In Range throwing error when no input numbers
- Fixed URL in the agent Rest API popup

## How to test:

## Additional information:

### DIFF
--- a/apps/docs/docs/api/index.md
+++ b/apps/docs/docs/api/index.md
@@ -59,7 +59,7 @@ The interface for an agent object that's based on the `agentSchema`.
 
 #### Defined in
 
-[packages/core/shared/src/schemas.ts:75](https://github.com/Oneirocom/Magick/blob/f79bad6b/packages/core/shared/src/schemas.ts#L75)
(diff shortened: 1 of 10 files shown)


Knar33 (approved):
Looks great!

-  logical operator now has valid options - still doesn't work (there's another issue for this that can be fixed later)
- Is Null or Undefined node is working now
- Get values from array output is properly labeled
- Number variable is correctly outputting numbers
- Cast to number correctly outputs a number instead of a string
- Text rule matcher minimum/maximum works properly
- Merge objects correctly handles both cases, passing an object or not passing an object
- In Range now works without input sockets defined
- The URL is now showing the agent id in the agent rest api settings window

Knar33 (approved):

PR #955, opened 2023-05-30
Fix delete and cleanup of agent when only one agent

lalalune:
## What Changed:
Right now the agent doesn't delete if the newAgent count === 0. This changes the order of logic to allow the deleteOldAgents function to run before the conditional check


Knar33 (approved):
I was able to replicate the bug on Development using a single Agent, and it didn't stop responding when I disabled it. 

I repeated the same steps in your branch and it went offline and stopped responding when I disabled it.  

Looks good!

PR #962, opened 2023-05-30
adds redis options for username and password

benbot:
## What Changed:

Added environment variables for redis.

## How to test:

Ensure that local works and that this runs in dev. be sure to set the environment vars in heroku


coffeeorgreentea:
Can we add defaults to the .env?

PR #981, opened 2023-06-05
Fixed agent window

FabriceIRANKUNDA:
## What Changed:

- fixed agent window public variables when root spell changed
- fixed combine text node

## How to test:

- Change the root spell's public variable
- revisit the agent window
- create combine text node and pass two the text variables to combine without setting delimiter
- press send button at playtest

## Additional information:


Knar33 (approved):
- public variables seem to update on the agent window
- combine text node properly handles an undefined delimeter

PR #968, opened 2023-06-01
Added Local models support

sshivaditya2019:
## What Changed:

Support for local models added using [Text-Generation-webui](https://github.com/oobabooga/text-generation-webui)

## How to test:

Choosing the model in GenerateText and CreateTextEmbeddings

## Additional information:

Updates Documents dependencies

PR #993, opened 2023-06-06
Add Support for text-generation-webui and local models

lalalune:
## What Changed:
This PR fixes the noted issues with https://github.com/Oneirocom/Magick/pull/968 and also changes some things that are related to it being copied from OpenAI. Also moves into packages/plugins folder ahead of that breaking change.


sshivaditya2019 (dismissed):
LGTM

benbot (changes requested):
I'm not too familiar with the text-generation-ui's API, but this looks good to me by reading it. Really hyped for this because I have some local models that are a pain to load by hand that i've been wanting to try out.

Just a few small changes.

We need to change the `console.log` statements to use our logger.
And this is all very specific to oobabooga's text-generation-ui, which is fine, but we should name it as such. So maybe like a `text-generation-ui-api` plugin instead of `localmodel`. One day we may just provide a way to load models ourselves, yknow?

Knar33:
> Yeah i can't seem to get the plugin loading so i don't see the local model options in the generatetext or create text embedding nodes :(

Did you add the "localmodel" to your plugins list in .env?

sshivaditya2019:
You need to run `npm install` after adding a new plugin.  After the plugin is added you should be able to see `Loaded Chat Model` in Generate Text Node

sshivaditya2019 (approved):

benbot (approved):

PR #995, opened 2023-06-07
Pino logger.error can't be called implicitly

benbot:
## What Changed:

Pino loggers can't be called implicitly (because it uses some magic)

Some errors were bubbling up in dev and we were getting crashes trying to call pino in this way

## How to test:

run server locally without redis being online
observe a TON of ERRCONNREFUSED errors, but the agent doesn't crash instantly


fire:
I think it's ok.

Needs a @Knar33 check if see if we missed anything. @benbot How can I cause the crash?

ultrabot2022:
it works on my side now

Issue #788, opened 2023-05-04
Is Variable True not testing numbers properly


Knar33:
### Magick version

4d81827

### System information

Ubuntu 22.04.1, firefox

### Issue description

When testing Is Variable True with numbers, it always returns false. 

If I enter a 1 I expect it to return true but instead it returns false. This could be an issue with the number variable node. If I put in "true" into the number variable node it returns true.

### Steps to reproduce

![image](https://user-images.githubusercontent.com/4692561/236334295-fad7c1bb-2fa3-4c74-bc95-c6c5c6046cad.png)

Create a diagram like above. When I enter anything into the chat window, I should get a response back echoing my input message because my number is set to 1 (true). I get a response if I hook up the false node, indicating that 1 is being evaluated as false.

Knar33:
This is still happening (current production build):

![image](https://github.com/Oneirocom/Magick/assets/4692561/c24f8ebf-759f-4d0f-8679-488b660b2120)

Knar33:
I opened up another ticket about the Is Variable True node so it might be a wider issue with the node. Also we may have fixed one thing (numbers outputting numbers) which caused this to regress.

PR #997, opened 2023-06-07
fix Is Variable True testing value 

Emmyn5600:
## What Changed:

I made small fix on isVariableTrue node file now , we can get a response if I hook up the true node not the false one,

## How to test:

create the same diagram described in #788  and try to play with it's chat window.

## Additional information:

the spell should only go through only when we have a value in chat window and we have also hood up the true node 

Screenshot from 2023-06-07 23-34-42

Screenshot from 2023-06-07 23-38-24


FabriceIRANKUNDA (commented):
LGTM! is variable node seems to work, you just need to resolve all index.md conflicts

coffeeorgreentea (approved):

PR #1006, opened 2023-06-09
Fixing deleting a spell while the spell is open and active

Emmyn5600:
## What Changed:

I made few changes on the onDelete function which handles spell delete , I first made sure that I close the spell before deleting it and I also made changes on the back button which is on open project form and I made sure that we don't use the browser history to navigate back , insead i navigate the user back to `/magic` page

## How to test:

- Create a spell
- Have that spell as your currently active spell tab in the spell editor
- Delete the spell (via the Open spell dialog)
- Confirm the deletion
- check if  the spell is still open in your editor

### DIFF
--- a/packages/editor/src/screens/HomeScreen/AllProjects.tsx
+++ b/packages/editor/src/screens/HomeScreen/AllProjects.tsx
@@ -137,7 +137,7 @@ const AllProjects: React.FC<AllProjectsProps> = ({
         <FileInput loadFile={loadFile} />
         <Button
           onClick={() => {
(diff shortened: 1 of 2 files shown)


FabriceIRANKUNDA (changes requested):
Deleting a spell which is not in active tabs will not be deleted

coffeeorgreentea (approved):

coffeeorgreentea (approved):

coffeeorgreentea (approved):

Knar33 (approved):
Looks good!

benbot (approved):

PR #1008, opened 2023-06-09
Fix wrong direction of some VRM models 

ultrabot2022:
## What Changed:
fixed some vrm models direction.

## How to test:
drag/drop any vrm models on the avatar page.


benbot (approved):
Code looks good to me

Knar33 (approved):
The VRMs I had the issue with work now

fire:
For future note:

1. this works as a workaround / hack
2. As far as I know when we start playing animations we need to apply the rotation to the entire vrm scene.

PR #1011, opened 2023-06-10
Fix string matcher

lalalune:
## What Changed:
String matcher was changed and not tested for both conditions, so it's currently broken, this reverts the conditional to fix it.


FabriceIRANKUNDA (commented):
Looks good!

benbot (approved):

PR #1020, opened 2023-06-12
dev to prod

benbot:


fire (approved):
Where's the how to test section? 👼

> BenBot — Today at 07:15
> No test for this. We're just moving our development branch into prod
> so no changes, just releasing the code to our alpha testers now that it soaked all weekend without canary dying

PR #1022, opened 2023-06-12
Refactor requests window

Emmyn5600:
## What Changed:

I refactored the request table by add the `TableComponent` reusable component to display all requests available 

## How to test:

if you  already have some request data , the go to the requests window , if not create a new discord spell and then interact with is in the chat window by sending few requests and the go back to the request window to view all the requests you have sent . there you can even delete a request by click on action.

## Additional information:
Screenshot from 2023-06-12 20-01-13
Screenshot from 2023-06-12 20-01-20

### DIFF
--- a/packages/editor/src/screens/RequestWindow/RequestTable.tsx
+++ b/packages/editor/src/screens/RequestWindow/RequestTable.tsx
@@ -110,6 +110,7 @@ function RequestTable({ requests, updateCallback }) {
     // close the action menu
     handleActionClose()
     updateCallback()
+    setSelectedRows([]);
   }
 
     // Handle events deletion


Knar33 (dismissed):
Seems to work for me. it's a bit ugly when looking at embeddings though, not sure a good way to fix it just figured I'd point it out:
image

FabriceIRANKUNDA:
> Seems to work for me. it's a bit ugly when looking at embeddings though, not sure a good way to fix it just figured I'd point it out: image

I think in this case we need to collapse the embeddings and show them up when expanded, @Emmyn5600 you can view the Figma design and use the collapse to handle this case.

FabriceIRANKUNDA (approved):

coffeeorgreentea (approved):

packages/core/server/src/services/requests/requests.class.ts
```diff
@@ -24,7 +24,29 @@ export type RequestParams = KnexAdapterParams<RequestQuery>
...
+> extends KnexService<Request, RequestData, ServiceParams, RequestPatch> {
+
+ /**
+   * Remove Requests.
+   * This function removes Requests from the database.
+   * @param {string[]} id - The ID of the Requests to remove.
+   * @returns {Promise<any>} - The removed Requests data.
+   */
+  // @ts-ignore
+  async remove(id: string): Promise<any> {
```
benbot, line 36: We're overriding the default `remove` function here. 
I don't think we want to do that. It'd be better if we had a `removeMany` function that took an array of id strings.

packages/core/server/src/services/requests/requests.class.ts
```diff
@@ -24,7 +24,29 @@ export type RequestParams = KnexAdapterParams<RequestQuery>
...
+    const ids = id.split('&')
```
benbot, line 37: It's not clear that we would need to pass in a string of '&' separated ids. This is why i'd think a `removeAll` function would be best.

packages/core/server/src/services/requests/requests.class.ts
```diff
@@ -24,7 +24,29 @@ export type RequestParams = KnexAdapterParams<RequestQuery>
  */
 export class RequestService<
   ServiceParams extends Params = RequestParams
-> extends KnexService<Request, RequestData, ServiceParams, RequestPatch> {}
+> extends KnexService<Request, RequestData, ServiceParams, RequestPatch> {
+
+ /**
+   * Remove Requests.
+   * This function removes Requests from the database.
+   * @param {string[]} id - The ID of the Requests to remove.
```
benbot, line 32: this docstring is wrong :see_no_evil:

benbot (changes requested):

packages/editor/src/screens/RequestWindow/RequestTable.tsx
```diff
@@ -50,157 +47,53 @@ const GlobalFilter = ({ globalFilter, setGlobalFilter }) => {
...
+  const handleActionClose = () => {
+    setAnchorEl(null)
+    setSelectedRow(null)
   }
 
   // Handle request deletion
   const handleRequestDelete = async (event: any) => {
-    const resp = await fetch(`${API_ROOT_URL}/request/${event.id}`, {
-      method: 'PATCH',
+    const resp = await fetch(`${API_ROOT_URL}/request/${selectedRow.id}`, {
```
benbot, line 95: I think this is the wrong endpoint. you may be able to use the feathers client instead of fetch here.

something like

`await app.service('requests').removeMany([requestIds])`

benbot (approved):

FabriceIRANKUNDA (changes requested):
- Delete many still does not work for me, look at the screenshot below in the console there is a feathers' error
- Also when I selected a row and delete it using the delete option on the menu, the selected row state should be cleared because if it stays and I press the delete selected button a request to delete a row which has been already deleted can be triggered to an error!

Screenshot 2023-06-20 at 08 56 47

FabriceIRANKUNDA (approved):

FabriceIRANKUNDA:
> * Delete many still does not work for me, look at the screenshot below in the console there is a feathers' error
> * Also when I selected a row and delete it using the delete option on the menu, the selected row state should be cleared because if it stays and I press the delete selected button a request to delete a row which has been already deleted can be triggered to an error!
> 
> Screenshot 2023-06-20 at 08 56 47

LGTM! npm run dev fixed it!

benbot (approved):

PR #1024, opened 2023-06-12
Add import (replace) for project

lalalune:
## What Changed:
Right now when you import a project it adds on top of what you already had. For handling big projects across machines, etc, I have found it's much more useful to delete everything first before importing the project. This changes Project window Import to have an Import (Add) and Import (Replace) function


packages/core/client/src/components/Drawer/ProjectWindow/index.tsx
```diff
@@ -70,33 +70,86 @@ const ProjectWindow = ({ openDrawer }) => {
...
+    await axios({
+      url: `${globalConfig.apiUrl}/agents`,
+      method: 'DELETE',
+      data: { projectId: globalConfig.projectId },
+      headers: {
+        Authorization: `Bearer ${token}`,
+      },
+    })
+    // make a call to the spell service to delete all spells
+    await axios({
```
benbot, line 133: Can we just use one of fetch, axios, or the feathers app?

Having 3 different ways of calling our api is not great :(

benbot (changes requested):

packages/core/client/src/components/Drawer/ProjectWindow/index.tsx
lalalune: I'm just copying the pattern that was already there. We can replace axios with fetch across the board, or go to service model, but it's beyond the scope of the feature and my ability to invest time into things that don't bring immediate value to users is very limited.

benbot: > I'm just copying the pattern that was already there. We can replace axios with fetch across the board, or go to service model, but it's beyond the scope of the feature and my ability to invest time into things that don't bring immediate value to users is very limited.

I see. This is true :(

I'll get it on the backlog

lalalune:
> Also can you rebase this on top of development to blow away the merge commit

No, I don't have time, sorry

lalalune:
> It seems to be working, but when I imported a project with no documents, the documents from my old project were still there. I switched between the two projects a few times and my documents kept duplicating.

Will look, should delete all documents, but only if there is a documents: [] object in thje project (but there should be)

I tried it and didn't have issues, but i had documents for both, so will check

benbot (dismissed):

benbot (approved):

PR #1025, opened 2023-06-12
Add some new nodes

lalalune:
## What Changed:
This PR adds some useful new nodes
- Extract From Array -- make a new array from the value of key inside the input array. So if you have a document and you just want an array of the content, values, for example
- Get Length -- get the length of string
- Split By Sentence -- Split a string down to a under a maximum length, but in a sentence. So if you want to write a tweet but you want each tweet to be coherent sentences (or just get the first array value for a single tweet)
- Trim -- Trim the whitespace at beginning and end of a string

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -178,6 +178,7 @@
         "rete-connection-plugin": "0.9.0",
         "rollup-plugin-ignore": "1.0.10",
         "rollup-plugin-node-polyfills": "0.2.1",
+        "sentence-splitter": "^4.2.1",
         "sepia-speechrecognition-polyfill": "1.0.0",
         "serp": "2.2.2",
         "serve-favicon": "2.5.0",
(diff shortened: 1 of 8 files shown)


Knar33:
- Extract from array works as described
- Get length works as described (including blank values)
- Split by sentence gave me some unexpected output. With the input "this is a sentence. And this is another sentence. And this is a really long sentence with more than 140 charachers blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah lorem ipsum etc etc." I get the output array ["this is a sentence."," ","And this is another sentence."," ","And this is a really long sentence with more than 140 charachers blah blah blah blah blah blah blah blah blah blah blah blah blah blah blah ","lorem ipsum etc etc."]. The last sentence being cut up makes sense but there are two empty string entries thrown in between the sentences that are short enough.
- Trim works as expected

benbot (dismissed):

benbot (approved):

PR #1026, opened 2023-06-12
Fix output

lalalune:
## What Changed:
Right now, "default" output works, but sending output to a specific out isn't working. This fixes it. Tested with taking a tweet input and outputting to a Discord channel by adding a channel ID to the event


Knar33:
Can you give me the spell you tested with? I got an error when I tried this 
image

benbot (approved):
It seems to work for me with no errors. Merging it in

PR #1027, opened 2023-06-13
Slow down zoom in/out

lalalune:
## What Changed:
This introduces a "zoom lerp factor" which slows down zoom by 1/4th. This is a temporary fix for an impossible zoom experience on Macbooks.

- Macbook only, so someone with a Mac will need to test. Shouldn't do anything on a PC.

- There is a little glitch at the beginning. To fix this, we would need to update the Oneirocom rete.js fork and handle this zoom internally, seems to be fixed?

- Ideally could control zoom sensitivity from UI, but this sets it on constructor.

- Ideally isn't in the zoom restrictor, although that nicely handles the background etc as well for now.

### DIFF
--- a/packages/core/client/src/components/Drawer/ProjectWindow/index.tsx
+++ b/packages/core/client/src/components/Drawer/ProjectWindow/index.tsx
@@ -98,6 +98,47 @@ const ProjectWindow = ({ openDrawer }) => {
   }
 
   /**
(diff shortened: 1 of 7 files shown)


FabriceIRANKUNDA (approved):
Big improvement on zoom in AIDE

PR #1040, opened 2023-06-15
join/leave discord VC nodes working, list text/voice channels

lalalune:
## What Changed:
This PR enables voice functionality

- Join voice node
- Leave voice node
- List Voice Channels node
- List Text Channels node

Does voice actually work? Well, no. We need TTS and STT handler plugins so that we can set voices dynamically from these. But the agent joins and leaves via channel properly.


Knar33:
I can get the bot to list the voice and text channels, but I can't get it to join a voice channel. If I try the channel ID I get no error but it doesn't join the channel, and when I send the channel name I get 
```
ERROR RUNNING SPELL, {"requestBody":{},"rawError":{"code":50035,"errors":{"channel_id":{"_errors":[{"code":"NUMBER_TYPE_COERCE","message":"Value \"General\" is not snowflake."}]}},"message":"Invalid Form Body"},"code":50035,"status":400,"method":"GET","url":"https://discord.com/api/v10/channels/General"}
```

benbot (approved):

PR #1041, opened 2023-06-15
Restrict to VRM1 and ignore forward fixing

ultrabot2022:
## What Changed:
we can load only VRM1 files for now

## How to test:
drag/drop any vrm files. if it is vrm0 or invalid file, then we will show the default avatar.

### DIFF
--- a/packages/plugins/avatar/client/src/hooks/useVrm.ts
+++ b/packages/plugins/avatar/client/src/hooks/useVrm.ts
@@ -24,7 +24,9 @@ export const useVrm = (vrmUrl) => {
         logger.error('non standard model %s', vrmUrl)
         gltf = await gltfLoader.loadAsync(DEFAULT_MODEL)
       }
-      if (!gltf) {
+
+      //ignore vrm0 file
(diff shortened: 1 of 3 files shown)


packages/plugins/avatar/client/src/hooks/useVrm.ts
```diff
@@ -24,7 +24,9 @@ export const useVrm = (vrmUrl) => {
         logger.error('non standard model %s', vrmUrl)
         gltf = await gltfLoader.loadAsync(DEFAULT_MODEL)
       }
-      if (!gltf) {
+
+      //ignore vrm0 file
+      if (!gltf || gltf.userData.vrm.lookAt.faceFront.z <= 0) {
```
fire, line 29: Use a toast to report error.

fire: Tracked in https://github.com/Oneirocom/Magick/issues/1048

benbot (approved):

fire (approved):
Looks good. Removes rotation fix. Blocks vrm0. 

Probably want a QA check, but looks good.

PR #1055, opened 2023-06-16
Fix a bunch of typescript issues

michaelsharpe:
## What Changed:

- fix a bunch of typescript errors.  Some of them have been ts-ignored, however all have comments, and some have a todo on them. Some are not worth trying to fix and might be easier to deal with upgrading the packages they come from (like react-table)

## How to test:

Ensure things still work.  Lol.

## Additional information:


michaelsharpe:
Adding some more, but cant figure out how to move it back to "in draft"

benbot (approved):

PR #1057, opened 2023-06-18
Fix plugin reloading by synchronously loading plugin imports

lalalune:
## What Changed:
Right now plugins are async loading into agent and server. The problem is that they don't rebuild on reload. This was fixed before by forcing the plugins to be implicit dependencies, however this would require all plugins to be added. Synchronous plugin loading fixes the rebuild.

### DIFF
--- a/apps/agent/project.json
+++ b/apps/agent/project.json
@@ -6,9 +6,7 @@
   "implicitDependencies": [
     "@magickml/server-core",
     "@magickml/core",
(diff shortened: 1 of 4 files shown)


lalalune:
This fixed the problem of the plugins not updating entirely, but didn't change plugins not updating on refresh. I believe the solution is to manually put in the plugin dependencies... but IDK what happens if the dependencies are unused :-/

PR #1089, opened 2023-06-23
Add function calling in OpenAI chat completions

lalalune:
## What Changed:
This PR adds a "Function" input which can be passed along with the message. If a function is returned, it will be returned as the result instead of the message. Functions are in JSON string format and will need to be parsed to an object on response.


benbot (approved):

benbot (approved):

fire:
One of the two dependencies for https://github.com/Oneirocom/Magick/issues/1124

🎉

PR #1094, opened 2023-06-23
fixing page autochange on events pages

Emmyn5600:
## What Changed:

I added initial state on useTabel hook with contains pageIndex attribute which will help us to make sure that if we refresh our page after deleting items , that same page will be returned back , instead of loading the first page which causes items reorder 

## How to test:

1. Add many events on events table
2. go to any page other than the first one 
3. try deleting any item 

## Additional information:

[Screencast from 23.06.2023 16:15:36.webm](https://github.com/Oneirocom/Magick/assets/55635977/bd5debdd-3232-45bf-a00a-6eebe72b6a92)


FabriceIRANKUNDA (commented):
Looks good! just remember to add the logic of navigating to the previous page on delete many as well

benbot (approved):

PR #1099, opened 2023-06-24
Update elevenlabs plugin

lalalune:
## What Changed:
Changes the Text To Speech node to cache a file and pass around a URI for it. Changes text2speech to textToSpeech.

### DIFF
--- a/packages/plugins/elevenlabs/server/src/functions/textToSpeech.ts
+++ b/packages/plugins/elevenlabs/server/src/functions/textToSpeech.ts
@@ -3,7 +3,7 @@ import {
   CompletionHandlerInputData,
 } from '@magickml/core'
 import axios from 'axios'
-
+import { v4 as uuidv4 } from 'uuid'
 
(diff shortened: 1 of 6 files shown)


benbot (approved):

packages/plugins/elevenlabs/server/src/functions/textToSpeech.ts
```diff
@@ -17,12 +17,25 @@ async function callTextToSpeechApi(text: string, voice_id: string, stability: nu
...
+    responseType: 'stream'
   });
 
-  return response.data;
+
+  const fs = await import ('fs')
+  const uuid = uuidv4()
+ 
+  // write response.data to files/
+  const writer = fs.createWriteStream(`files/${uuid}.mp3`)
```
benbot, line 32: Can't we just provide a stream instead of storing a file?

fire:
@benbot Has the design of this changed now that we propose to upload binaries to a cloud service?

michaelsharpe:
@benbot @coffeeorgreentea this might be useful as we were talking about getting eleven labs working properly.

PR #1117, opened 2023-06-26
Fixing page auto-change on documents and requests windows

Emmyn5600:
## What Changed:

I added the initial state on useTable hook with contains pageIndex attribute which will help us to make sure that if we refresh our page after deleting items, that same page will be returned back, instead of loading the first page which causes items to reorder

## How to test:

1. Add many requests/documents on request or document table
2. Go to any page other than the first one
3. Try to delete any item

## Additional information:

[Screencast from 26.06.2023 17:57:42.webm](https://github.com/Oneirocom/Magick/assets/55635977/9e5733d5-0d70-403a-8d03-232bdddef336)


FabriceIRANKUNDA (commented):
LGTM!

FabriceIRANKUNDA (approved):

PR #1132, opened 2023-06-29
Clean up package.json, update wagmi, remove unused deps

lalalune:
## What Changed:
NPM is throwing a lot of warnings about security issues. Some of this will mean going in and pinning dependencies and tracking down each thing. Some of it comes from just having unused features and package bloat.

This PR does the following
- Update wagmi to the latest version, which requires changing some loading code. This is specific to the ethereum client, and tbh the wagmi provider should all go into the plugin and not otherwise be loaded (future PR)
- Remove capacitor and electron. Eventually this would be good to put back in, but it's a lot of bloat, nobody else knows how it works and it's not on the cloud path
- Remove node polyfills from vite, which were causing some funky issues with newer version of nx
- Remove unused packages in package.json
- Update some packages

I tried updating to nx 16.4.0 but there is some different behavior / issues, so incremented only up to 16.3.1

End result shouldn't be any different. Project should be slightly more lightweight.


fire:
I checked the capacitor and electron bits a while ago and they're good to remove.

benbot (approved):

PR #1165, opened 2023-07-07
Add tooltips to all left nav buttons 

Emmyn5600:
## What Changed:

I wrapped a Tooltip component around the DrawerItem component and what we have to do is to only pass a tooltip text description as prop and it will be displayed when user hovers over an element 

## How to test:

Hover over sidebar menu items in the AIDE 

## Additional information:

[Screencast from 07.07.2023 15:09:02.webm](https://github.com/Oneirocom/Magick/assets/55635977/b39f3596-8813-4a57-aa1d-190ce30cd1ec)


FabriceIRANKUNDA (commented):
You can bring the tooltip close to the item I think!

benbot (approved):

PR #1169, opened 2023-07-10
Improved Document Chunking and document table UI

sshivaditya2019:
## What Changed:

- Improved document chunking (2500 words with 10% sliding window)
- Full text search node for extracting relevant paragraphs by entities

## How to test:

Test Spell with EntityExtraction subspell

## Additional information:

gpt3.5turbo supports a maximum of 5 documents

### DIFF
--- a/packages/core/server/src/services/documents/documents.class.ts
+++ b/packages/core/server/src/services/documents/documents.class.ts
@@ -69,33 +69,44 @@ export class DocumentService<
    * @return {Promise<any>} The found documents
    */
   async find(params?: ServiceParams): Promise<any> {
-    const db = app.get('dbClient')
+    const db = app.get('dbClient');
     if (params.query.embedding) {
-      const param = params.query
+      const param = params.query;
+      const keywords = param.content ? param.content.split(' ') : [];
+  
       const querys = await db('documents')
         .select('*')
-        .where({
-          ...(param.type && { type: param.type }),
-          ...(param.id && { id: param.id }),
-          ...(param.sender && { sender: param.sender }),
-          ...(param.client && { client: param.client }),
-          ...(param.channel && { channel: param.channel }),
-          ...(param.projectId && { projectId: param.projectId }),
-          ...(param.content && { content: param.content }),
+        .where((builder) => {
+          builder.where({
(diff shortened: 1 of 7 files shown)


packages/core/server/src/services/documents/documents.class.ts
```diff
@@ -69,33 +69,52 @@ export class DocumentService<
...
-        .where({
-          ...(param.type && { type: param.type }),
-          ...(param.id && { id: param.id }),
-          ...(param.sender && { sender: param.sender }),
-          ...(param.client && { client: param.client }),
-          ...(param.channel && { channel: param.channel }),
-          ...(param.projectId && { projectId: param.projectId }),
-          ...(param.content && { content: param.content }),
+        .where((builder) => {
+          if (param.type) {
```
benbot, line 80: something like `param.type && bulder.where(...)` might be easier to read

packages/editor/src/screens/DocumentWindow/DocumentTable.tsx
```diff
@@ -310,6 +349,16 @@ const handleSave = async (selectedModel) => {
               >
                 Refresh
               </Button>
+              {/* <Button
```
benbot, line 352: can we remove this if we don't need it?

packages/editor/src/screens/DocumentWindow/DocumentTable.tsx
```diff
@@ -190,6 +195,35 @@ function DocumentTable({ documents, updateCallback }) {
     setSelectedRow(null)
   }
 
+  // Handle multi document deletion
+  const handleDeleteMany = async (event: any) => {
+    const ids = selectedRows.join('&');
+    const isDeleted = await fetch(`${API_ROOT_URL}/documents/${ids}`, {
```
benbot, line 201: We're trying to unify how we fetch data from the api. could you try using the feathers client from the `useFeathers` hook?

if you do a text search for `app.service('documents')`  you'll find some examples in the AIDE

benbot (changes requested):

michaelsharpe:
@benbot  is this still useful to us? Maybe @Knar33 would have some context?

PR #1172, opened 2023-07-11
Integrating of Chatbox

ultrabot2022:
## What Changed:
A ChatBox has been added on the AIDE


benbot (changes requested):
We should extract these changes into it's own "ChatBox" component that we can import and use on other pages as well.

benbot (approved):

PR #1173, opened 2023-07-11
Openmeter

coffeeorgreentea:
## What Changed:

Metering for openai and googleai models.


packages/plugins/openai/server/src/index.ts
```diff
@@ -16,6 +16,7 @@ import {
   makeTextCompletion,
   makeTextEmbedding,
 } from './functions'
+import { trackUsage } from '../../../../core/server/src/metering/openaiMetering'
```
benbot, line 19: this should be imported by the package name `@magickml/server-core`

packages/core/server/src/metering/openaiMetering.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * Tracks the usage of the OpenAI API by sending a metering event.
+ *
+ * @param {Function} handler - The handler function.
+ * @returns {Function} - The decorated handler function.
+ */
+export function trackUsage(handler: Function) {
+  console.log('Tracking usage of OpenAI API')
```
benbot, line 46: pino logger please <3

packages/plugins/openai/server/src/index.ts
```diff
@@ -27,9 +28,9 @@ const { secrets } = shared
  */
 const completionHandlers = {
   text: {
-    text: makeTextCompletion,
-    chat: makeChatCompletion,
-    embedding: makeTextEmbedding,
+    text: trackUsage(makeTextCompletion),
```
benbot, line 31: In this case, it may make more sense to just have trackUsage be a normal function that you call from with in the other `makeWhateverCompletion` functions.

I was thinking we could use decorators at the plugin level to wrap completion handlers, but if we're just going to have them on a per plugin basis, might as well just make them regular functions.

packages/config/src/config.ts
```diff
@@ -100,3 +100,5 @@ export const bullMQConnection = {
 }
 
 export const PINO_LOG_LEVEL = getVarForEnvironment('PINO_LOG_LEVEL') || 'info'
+
+export const OPENMETER_ENDPOINT = getVarForEnvironment('OPENMETER_ENDPOINT') || 'http://localhost:8888'
```
benbot, line 104: I don't think we should have this default to anything, but if OPENMETER_ENDPOINT is undefined, we just log a warning and skip metric reporting.

This way users who run magick locally don't _need_ to have openmeter set up

benbot (changes requested):

packages/config/src/config.ts
coffeeorgreentea: i have an openmeter enabled flag too now

packages/core/server/src/metering/openaiMetering.ts
coffeeorgreentea: will do

benbot (approved):

PR #1199, opened 2023-07-20
Intent detection

Knar33:
Added intent detection node


benbot (commented):
The changes look good, but does this mess with the documents page / nodes?

If metadata is null or undefined, how does the query work.

Maybe we can change the service to only add that portion of they query if we pass in a metadata specifically. If we don't then we just get all documents.

benbot (approved):

PR #1221, opened 2023-08-02
add command parser node

coffeeorgreentea:
## What Changed:

Adds a Command Parser that parses commands such with an optional string array of valid commands.
/command arg1 arg2
into outputs:
command: 'command'
args: ['arg1','arg2']
isCommand: boolean

## How to test:

Use the node

### DIFF
--- a/packages/core/shared/src/nodes/flow/CommandParser.ts
+++ b/packages/core/shared/src/nodes/flow/CommandParser.ts
@@ -73,7 +73,6 @@ export class CommandParser extends MagickComponent<Promise<WorkerReturn>> {
     )
 
     const triggerOutput = new Rete.Output('trigger', 'Trigger', triggerSocket)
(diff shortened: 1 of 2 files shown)


packages/core/shared/src/nodes/flow/CommandParser.ts
```diff
@@ -0,0 +1,154 @@
...
+  parseCommand(
+    input: string,
+    checkList: string[],
+    useCheckList: boolean
+  ): ParsedCommand {
+    let isCommand = false
+    let command = ''
+    let args: string[] = []
+
+    if (input.startsWith('/')) {
```
benbot, line 122: I have seen some bots use `!` instead of `/`

should we let people choose their own prefix?

coffeeorgreentea: yeah i like that

benbot (approved):

PR #1231, opened 2023-08-09
Quiets logs and fixes agent startup on server cold start

benbot:
## What Changed:

Changed the log level of heartbeats to "trace"

Agent startup messages on server cold start now contain timestamps. This ensures that we won't drop jobs if a job id matches a previous one.

## How to test:
Enable an agent then restart the server

Observe that the agent started

IMPORTANT: This will always work the first time, it's important to check a 2nd time so...

restart the server again and observe that the same enabled agent started up again.


michaelsharpe:
Looks good to me!

PR #1234, opened 2023-08-11
New AgentMenu Integration

Emmyn5600:
## What Changed:

Pulled agent data ....

## How to test:

Open up any spell and then start interacting with the top agent menu by creating new agent if there isn't any and selecting which one to use etc..

## Additional information:

https://github.com/Oneirocom/Magick/assets/55635977/09750529-1d6e-4351-af55-c4f6174881ce


FabriceIRANKUNDA (commented):
Creating a new agent from the agent page requires a refresh for it to show up in the agent drop-down, may be we can this in and fix that in a new PR.

coffeeorgreentea (approved):

PR #1241, opened 2023-08-16
Adding scroll to settings and change document title and type to group

Emmyn5600:
## What Changed:

changed settings editor class styles and changed word document to Knowledge in both spell work space and documents window

## How to test:

1. open any spell 
2. navigate to setting window and check if there is a scroll effect
3. Go to documents window and check , the word "document" was changed to "Knowledge"
4. Navigate to spell window and try add adding Knowledge nodes eg: Get Knowledge node or Store Knowledge node

## Additional information:

https://github.com/Oneirocom/Magick/assets/55635977/ad458db4-24d6-44a3-952b-13bf3446ab5e


FabriceIRANKUNDA (changes requested):
The info should also be updated, I guess.
A node cannot be called Store Knowledge and the info says that it stores documents

FabriceIRANKUNDA (commented):
I think these should also be changed! There is also one node called Documents to Content which should be updated 

Screenshot 2023-08-23 at 09 45 30

FabriceIRANKUNDA (approved):

PR #1258, opened 2023-08-22
Text editor

FabriceIRANKUNDA:
## What Changed:

- Save spell difference when text editor data changed
- Publish text editor clear on tab change
- Auto save in text editor

## How to test:

- open more than one spells
- changed text editor data in one spell and keep switching between different spell tab to see if you lost data

## Additional information:

### DIFF
--- a/packages/editor/src/components/EventHandler.tsx
+++ b/packages/editor/src/components/EventHandler.tsx
@@ -7,7 +7,7 @@ import md5 from 'md5'
 
 import {
   useLazyGetSpellByIdQuery,
-  useSaveSpellMutation,
+  useSaveSpellMutation
 } from '../state/api/spells'
 import { useLayout } from '../contexts/LayoutProvider'
 import { useEditor } from '../contexts/EditorProvider'
@@ -29,6 +29,7 @@ const EventHandler = ({ pubSub, tab }) => {
   const { enqueueSnackbar } = useSnackbar()
 
   const [saveSpellMutation] = useSaveSpellMutation()
+
   // TODO: is this a bug?
   const [getSpell, { data: spell }] = useLazyGetSpellByIdQuery({
     spellName: tab.name.split('--')[0],
@@ -47,7 +48,7 @@ const EventHandler = ({ pubSub, tab }) => {
       id: tab.id,
       projectId: config.projectId,
     })
-  }, [config.projectId, getSpell, tab.id, tab.name])
+  }, [config.projectId, getSpell, tab.id, tab.name, tab])
 
   useEffect(() => {
     if (!spell) return
--- a/packages/editor/src/components/EventHandler.tsx
+++ b/packages/editor/src/components/EventHandler.tsx
@@ -48,7 +48,7 @@ const EventHandler = ({ pubSub, tab }) => {
       id: tab.id,
       projectId: config.projectId,
     })
-  }, [config.projectId, getSpell, tab.id, tab.name, tab])
+  }, [config.projectId, getSpell, tab])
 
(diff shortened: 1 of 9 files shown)


packages/editor/src/editor.ts
```diff
@@ -230,10 +230,10 @@ export const initEditor = function ({
 
   // Functions to load and run spells
   editor.loadSpell = async (spell: SpellInterface) => {
-    console.log('Loading spell in editor')
     if (!spell) return console.error('No spell to load')
     const _graph = spell.graph
     const graph = JSON.parse(JSON.stringify(_graph))
+    console.log('Loading spell in editor###########################', graph)
```
benbot, line 236: There are a few console.log statements here.

We should either remove them or make them `logger.debug` calls

packages/editor/src/contexts/InspectorProvider.tsx
```diff
@@ -66,26 +66,28 @@ const InspectorProvider = ({ children, tab }) => {
...
     return unsubscribe as () => void
-  }, [events, subscribe, publish])
+  }, [events, subscribe, publish, inspectorData])
 
   // Subscribe to text editor changes
   useEffect(() => {
     const unsubscribe = subscribe(
       events.$TEXT_EDITOR_SET(tab.id),
       (event, data) => {
+        console.log(":::::::::::::::::::::::::::", data)
```
benbot: here

packages/editor/src/contexts/InspectorProvider.tsx
```diff
@@ -66,26 +66,28 @@ const InspectorProvider = ({ children, tab }) => {
             control: control,
             options: control.options,
           }
+          console.log("}}}}}}}}}}}}}}}}}}}}}}}}", textData)
```
benbot: and here

packages/editor/src/components/EventHandler.tsx
```diff
@@ -47,7 +48,7 @@ const EventHandler = ({ pubSub, tab }) => {
       id: tab.id,
       projectId: config.projectId,
     })
-  }, [config.projectId, getSpell, tab.id, tab.name])
+  }, [config.projectId, getSpell, tab.id, tab.name, tab])
```
benbot, line 51: if `tab.id` and `tab.name` are being checked here already, do we need `tab`?

And if we do need `tab` here, then can we remove `tab.id` and/or `tab.name`

benbot (changes requested):

packages/editor/src/components/EventHandler.tsx
FabriceIRANKUNDA: > if `tab.id` and `tab.name` are being checked here already, do we need `tab`?
> 
> And if we do need `tab` here, then can we remove `tab.id` and/or `tab.name`

We do not need `tab.id` and `tab.name` since they are not triggering useEffect when tab changed, so `tab` is enough

FabriceIRANKUNDA: > if `tab.id` and `tab.name` are being checked here already, do we need `tab`?
> 
> And if we do need `tab` here, then can we remove `tab.id` and/or `tab.name`

We do not need `tab.id` and `tab.name` since they are not triggering useEffect when tab changed, so `tab` is enough

packages/editor/src/contexts/InspectorProvider.tsx
FabriceIRANKUNDA: > and here

I can't see this, may it is from the rebase you made!

benbot (approved):

PR #1252, opened 2023-08-21
Unstructured API

Knar33:
Changed Document upload to use unstructured.io API

PR #1259, opened 2023-08-22
Unstructured io ingestion

benbot:
Correction of #1252


package.json
```diff
@@ -69,6 +69,7 @@
     "@google-cloud/speech": "5.5.0",
     "@google-cloud/text-to-speech": "4.2.2",
     "@koa/cors": "4.0.0",
+    "@koa/multer": "^3.0.2",
```
benbot, line 72: can we remove multer

packages/core/server/src/services/documents/documents.class.ts
```diff
@@ -36,22 +40,38 @@ export class DocumentService<
...
-        modelName,
-        projectId: docData?.projectId,
-        secrets,
-      })
+    let elements = []
+    if (docData.content) {
+      elements = [
+        ...elements,
+        ...(await getUnstructuredData(
+          [{ text: docData.content, originalFilename: 'text.txt' }],
```
benbot, line 53: Is this originalFilename wrong?

package.json
benbot: koa multer is still here

coffeeorgreentea (approved):

PR #1283, opened 2023-09-03
potential typechat v2

coffeeorgreentea:
## What Changed:
DONT MERGE:  i want to add the old typechat node back in before we do
- Schema is moved into a fewshot control with typescript syntax highlighting (similar to text variable)
- Response type is moved to the inspector
- has a default value to show the user an example, and so that it works out of the box. (example ripped from here: https://github.com/microsoft/TypeChat/blob/main/examples/coffeeShop/src/coffeeShopSchema.ts)
- would break existing typechat spells :disappointed:

## How to test:

Use the typechat node and order a coffee

### DIFF
--- a/packages/core/shared/src/nodes/text/TypeChat.ts
+++ b/packages/core/shared/src/nodes/text/TypeChat.ts
@@ -14,6 +14,8 @@ import {
   MagickWorkerOutputs,
   WorkerData,
 } from '../../types'
+import { FewshotControl } from '../../dataControls/FewshotControl'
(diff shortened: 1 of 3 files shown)


michaelsharpe (approved):
Looks good!

coffeeorgreentea:
DONT MERGE:  i want to add the old typechat node back in before we do

michaelsharpe:
Any word on this? Would be awesome to get both nodes in there for users to use.

PR #1289, opened 2023-09-05
refactored agent public variables configuration logic

FabriceIRANKUNDA:
## What Changed:

- Refactored agent page's public variables

## How to test:

- Create spell with public nodes
- Set spell as a root spell to the agent
## Additional information:


coffeeorgreentea (dismissed):

coffeeorgreentea:
approved. we need to add logic for number public variables after this here and in cloud

FabriceIRANKUNDA:
> approved. we need to add logic for number public variables after this here and in cloud

@coffeeorgreentea current logic is handling number public variables ( as string), seems to work!

packages/core/client/src/components/Input/index.tsx
```diff
@@ -15,6 +15,8 @@ export const Input = ({
   onChange = e => {
     /* null */
   },
+  onKeyDown = e => {
```
benbot, line 18: is this supposed to do nothing?

packages/editor/src/screens/agents/AgentWindow/AgentPubVariables.tsx
```diff
@@ -55,35 +72,34 @@ const AgentPubVariables = ({
           marginBottom: '10px',
         }}
       >
-        {Object.values(publicVars).map((variable: any) => {
+        {publicVariables.map((variable, index) => {
```
benbot, line 75: Im always glad to get rid of an Object.values call :)

benbot (approved):

PR #1293, opened 2023-09-06
Task optimizations

michaelsharpe:
## What Changed:

Swicthed task plugin to use central store of tasks rather than hold them all in memory.  Moved our rete fork to internal so we can modify and have control of the source code.  Rete now lives in @magickml/rete.

## How to test:

Load a large spell, like Rob Lennons big ugly one, and watch your  system processes.  There should be no huge spikes in memory and things should remain stable.

Ensure agents work, and spells fire off nodes are light up the node graph.  This touches the task plugin, so should be well tested with a few different spells if possible.  Alternative, we push it to dev and test it there and patch as needed. This is a major fix needed to stabilize infrastructure.

## Additional information:


michaelsharpe:
@benbot lots of files touched and some formatting things also updated things.  Main files to pay attention to changes in are taskPlugin/index and taskPlugin/task.js.  Other changes to files were swapping to use internal rete rather than external.

PR #1295, opened 2023-09-06
fires iframe event in cloud for the back button

benbot:
## What Changed:

Fires iframe event to enable a back button in production.


coffeeorgreentea:
@benbot I changed !standalone to standalone

coffeeorgreentea (dismissed):

PR #1299, opened 2023-09-07
Espresso

coffeeorgreentea:
## What Changed:

## How to test:

## Additional information:


apps/cloud-agent-worker/src/plugins.ts
benbot, line 1: We should be ignoring this file

packages/core/server/src/services/Upload.class.ts
benbot, line 1: Does this break document upload?

benbot (commented):
I worry about removing the upload service. I thought that was used for unstructured document uploads (tho I could be wrong)

Other than that, I think a lot of these nodes should be moved into plugins

PR #1304, opened 2023-09-08
Fix spell manager updates

michaelsharpe:
## What Changed:

Fixes issue where spell running isn't activating nodes on the IDE graph.  Changes from graph not syncing to server.  Also fixes issue with fromSocket effecting the exclusive gate.

## How to test:

IDE should work normally.  new nodes added or removed should be synced to server and the whole graph should run properly when you activate it from the playtest.  Also exclusive gate should work.

Need to test if this fixes agents.

## Additional information:


benbot (approved):
I love this pr

PR #1307, opened 2023-09-08
moves from node to bun!

benbot:
## What Changed:

Moves from node to bun (bun.sh). It looks like more fun.


michaelsharpe:
Looks like we need to global install bun in a pre-build hook or something to ensure it is available.

PR #1313, opened 2023-09-11
Small fixes 1

michaelsharpe:
Merge after #1311 

## What Changed:

- Added local featyre flag for ENV to avoid using posthog locally.  Posthog actually overrides console logs and doest not give stack traces or proper line number references.
- Always log node errors to the IDE console even when the node is not told to log.

## How to test:

- Check if drawer doesnt show up when no env variable is set
-  add proper feature flag ENV variable to client env to ensure it can be turned on.

## Additional information:


michaelsharpe:
This logs the error message from the node to the IDE console. Requested by Dave. We were logging all outputs from all nodes which was noisy. And he often only cared about 1-2 nodes. But always cared about the error messages it surfaces. 

This change isn't adding anything that we didn't do before. It just gives more user control while maintaining visibility into the errors output by the nodes.

benbot (approved):

benbot (approved):

PR #1319, opened 2023-09-11
Fix exclusive gate and embeddings

michaelsharpe:
## What Changed:

- Exclusive gate should work as expected
- Fixed small bug in documents embedding with undefined content

## How to test:

- make sure exclusive gate works

## Additional information:

### DIFF
--- a/packages/core/shared/src/plugins/taskPlugin/task.ts
+++ b/packages/core/shared/src/plugins/taskPlugin/task.ts
@@ -125,11 +125,16 @@ export class Task {
     return value
   }
 
-  private filterNextTasks(con) {
+  private filterOutputConnections(con, fromNodeId: number | undefined) {
     const task = this.getTask(con.nodeId)
+    if (this.component.task.runOneInput && fromNodeId) {
+      if (task.nodeId === fromNodeId) return true
+      if (task.outputData) return true
+      return false
+    }
     // if (task.outputData) return true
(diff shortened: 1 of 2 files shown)


benbot (approved):

packages/core/shared/src/plugins/taskPlugin/task.ts
```diff
@@ -162,7 +167,10 @@ export class Task {
       await Promise.all(
         this.getInputs('output').map(async key => {
           const inputPromises = (this.inputs[key] as MagickReteInput[])
-            .filter(this.filterNextTasks.bind(this))
+            .filter(con => {
+              this.filterOutputConnections.bind(this)
```
benbot, line 171: Should this be bound here? Will this happen multiple times?

Also will this do anything? Doesn't bind return a function to be called?

michaelsharpe: Seems to work. The bund was necessary because we access this inside the bound function. Probably more elegant to just use .call instead and pass this in that way.

benbot: It seems better with call, but I think it would've worked without `bind`ing or `call`ing.

Bind returns a function, but you weren't using the value it returned, then just calling the normal not bound function

benbot (approved):

PR #1321, opened 2023-09-12
Fix avatar loading issue

ultrabot2022:
## What Changed:
Three.js version has been changed as 0.150 to fix issue of VRM1 file on the avatar page


package.json
```diff
@@ -190,7 +190,7 @@
     "styled-components": "5.3.11",
     "swr": "2.1.5",
     "tesseract.js": "4.0.6",
-    "three": "^0.154.0",
+    "three": "^0.150.0",
```
benbot, line 193: I'd there a way we can fix this without lowering the three js version?

This issue will just pop up again when we upgrade otherwise

ultrabot2022: I am checking multiple dependencies in different libraries of latest version of three.js

PR #1326, opened 2023-09-13
Display event in reverse chronological order

Emmyn5600:
## What Changed:

Added a function with uses swap algorithm to reverse events array and display them from the recent last one 

## How to test:

1. Create a discord bot 
2. interact with it in spell text input
3. navigate to events window and check if the recent request you made is the one on the top on of events table

## Additional information:


FabriceIRANKUNDA (commented):
In our backend, we're currently handling ordering events by date in descending order. However, there's a line of code that reverses the query result:
`
if (!param.embedding) {
    res.reverse();
}`

I believe we might be able to remove this line of code, but it's possible that @knar has insights into why it's there in the first place.

FabriceIRANKUNDA (approved):

PR #1360, opened 2023-10-12
Agent api overrides

michaelsharpe:
## What Changed:

Adds support for API calls to add properties to the event to they can set the sender, the client, the channel, etc.  Also adds support for passing through public variables, and secret overrides.

## How to test:

You can add these properties to a rest request, or try using the cloud PR here to send the properites via REST:

https://github.com/Oneirocom/Cloud/pull/495

## Additional information:

### DIFF
--- a/packages/plugins/rest/server/src/services/agentHttp/agentHttp.class.ts
+++ b/packages/plugins/rest/server/src/services/agentHttp/agentHttp.class.ts
@@ -1,3 +1,4 @@
+import { agent } from './../../../../../../core/server/src/services/agents/agents'
 // DOCUMENTED
 /**
  * For more information about this file see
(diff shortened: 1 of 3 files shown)


packages/plugins/rest/server/src/services/agentHttp/agentHttp.class.ts
```diff
@@ -65,6 +65,67 @@ const getAgent = async (
   return agent as unknown as Agent
 }
 
+const formatRequest = async (method, agentId, data, params) => {
```
benbot, line 68: Could we name this something like `formatRunSpellArgs`?

Might just make it a bit clearer as to what it's doing.

packages/plugins/rest/server/src/services/agentHttp/agentHttp.class.ts
```diff
@@ -65,6 +65,67 @@ const getAgent = async (
...
+    typeof agent.publicVariables === 'string'
+      ? JSON.parse(agent.publicVariables)
+      : agent.publicVariables
+  method = method.toUpperCase()
+
+  return {
+    agent,
+    spellId,
+    inputs: {
+      [`Input - REST API (${method})`]: {
```
benbot, line 104: we probably should validate that the method is one of POST, PUT, PATCH, GET, or DELETE. I think otherwise the spell would just silently not run if an invalid method is passed in

benbot (changes requested):

PR #1364, opened 2023-10-23
Basic document queue

michaelsharpe:
## What Changed:

- new queue and worker added in to process documents
- events emitted from the document service on progress and completion

## How to test:

- upload documents and ensure they are processed.  I will be adding in some in IDE progress reporters to the UI.

## Additional information:

Need to decide if we break this into a microservice or not.


benbot (changes requested):
This looks good for the successful case, but we should also have a dead letter queue for tracking and retrying failed jobs as well.

PR #1369, opened 2023-10-28
Palm api fallback

michaelsharpe:
## What Changed:

Improves a few error logs from spell runner and spell manager
Adds a fallback palm api key to all palm completions for now.

## How to test:

## Additional information:

### DIFF
--- a/packages/core/shared/src/spellManager/SpellRunner.ts
+++ b/packages/core/shared/src/spellManager/SpellRunner.ts
@@ -269,7 +269,7 @@ class SpellRunner {
   }
   error(message: string, error: unknown | null = null) {
     this.busy = false
-    this.logger.error(message, error)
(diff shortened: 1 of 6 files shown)


packages/core/shared/src/spellManager/SpellRunner.ts
```diff
@@ -375,7 +377,7 @@ class SpellRunner {
       // A run shouldnt take this long.  This is a hacl but we are replacing all this soon.
       setTimeout(() => {
         this.busy = false
-      }, 10000)
```
benbot, line 378: Should this be an environment variable ?

benbot (dismissed):
Looks good, but it'd be nice to have that timeout configurable

benbot (approved):

PR #1390, opened 2023-11-08
Subspell fixes

michaelsharpe:
## What Changed:

- Adds two new nodes - Socket Input and Socket Output
- Hides the subspells Event Input nodes from the parent if the event input is a subspell
- Adds an event socket to the Spell component
- Event socket passes down and out of subspell event input 
- Socket Input and Socket output both add sockets into subspell parent
- Socket Input and Socket output can both have default values put into them (only support strings right now)
- Parent socket names update when name of child socket input/output changes

## How to test:

## Additional information:

- Might want to add public variables to the parent spell sockets too


packages/core/shared/src/nodes/io/Spell.ts
```diff
@@ -333,9 +339,13 @@ export class SpellComponent extends MagickComponent<
     const { module, spellManager, app, agent } = _context
     const { secrets } = module
 
+    // We want to trigger off the subspell with the incoming event as its payload.
+    formattedInputs['Input - Subspell'] = inputs.event[0]
+    delete formattedInputs.event
```
benbot, line 344: Why `delete` instead of set to `null`?

`deleting` it makes it undefined, but `null` would let us know later on that you specifically removed that value.

packages/core/shared/src/plugins/modulePlugin/index.ts
```diff
@@ -165,14 +165,27 @@ function install(
           ) => {
             const modules = moduleManager.modules || []
             const currentNodeModule = node.data.spellId as string
-            if (!modules[currentNodeModule] && !graphData) return
+
+            // This is SUPER hacky but works as we store all editors on this.  Just need to see if it is kept up to date.
```
benbot, line 169: Could you expand on this commend explaining what's effectivly going on here and why?

The code isn't easy to grok on it's own :sweat:

packages/core/shared/src/plugins/modulePlugin/index.ts
```diff
@@ -205,7 +218,10 @@ function install(
 
           component.builder = async node => {
             // @ts-ignore
-            if (!component.noBuildUpdate) component.updateModuleSockets(node)
+            console.log('UPDATING SOCKETS!')
```
benbot, line 221: console log

packages/core/shared/src/plugins/socketGenerator/index.ts
```diff
@@ -53,6 +53,8 @@ function install(editor: IRunContextEditor) {
         nodeInputs.forEach(socket => {
           // If the input key is already on the node, return
           if (inputMap[socket.socketKey]) return
+          console.log('SOCKET', socket)
```
benbot, line 56: logz

packages/editor/src/App.css
```diff
@@ -93,11 +93,11 @@ body,
 .socket.output.trigger,
 .socket.input.trigger,
 .expanding.trigger {
-  background: var(--purple);
+  background: white;
```
benbot, line 96: Should we make new css variables for these values?

packages/editor/src/editor.ts
```diff
@@ -135,7 +136,7 @@ export const initEditor = function ({
   editor.use(DebuggerPlugin)
   editor.use(MultiCopyPlugin)
   editor.use(SocketGeneratorPlugin)
-  editor.use(MultiSocketGenerator)
+  // editor.use(MultiSocketGenerator)
```
benbot, line 139: Can we remove this if it's unused now?

packages/editor/src/windows/PlaytestWindow.tsx
```diff
@@ -188,21 +189,24 @@ const Playtest = ({ tab }) => {
   // Print to console callback function.
   const printToConsole = useCallback(
     (_, _text) => {
+      console.log('Playtest received!', _text);
       // check if _text is a string
-      if (typeof _text !== 'string')
-        return console.warn('could not split text, not a string', _text)
-      const text = `Agent: ` + _text
+      if (typeof _text !== 'string') {
+        console.warn('Could not split text, not a string', _text);
```
benbot, line 195: :eyes:

benbot (changes requested):

Knar33:
I tested everything and it all works. Only two notes:

- The old "Input - Default" sockets are confusing and should probably just be removed, even if that breaks backwards compatibility with old spells. I couldn't get it to even work
- The Run Spell by Name node needs an Outputs object (to pass the socket outputs out)

PR #1400, opened 2023-11-16
Stop event store and recall from lowercasing things

michaelsharpe:
## What Changed:

stop lowercasing events and treat inspector data same as socket data for event type

## How to test:

## Additional information:

### DIFF
--- a/packages/core/shared/src/nodes/events/EventRecall.ts
+++ b/packages/core/shared/src/nodes/events/EventRecall.ts
@@ -181,7 +181,9 @@ export class EventRecall extends MagickComponent<Promise<InputReturn>> {
 
       return events
     }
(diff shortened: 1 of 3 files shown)


packages/core/shared/src/nodes/events/EventRecall.ts
```diff
@@ -181,7 +181,7 @@ export class EventRecall extends MagickComponent<Promise<InputReturn>> {
 
       return events
     }
-    const typeSocket = inputs['type'] && inputs['type'][0]
+    const typeSocket = (inputs['type'] && inputs['type'][0]) as string
```
benbot, line 184: should this be `string | undefined`

benbot (approved):

PR #1422, opened 2023-12-11
Slack+Credential Manager

coffeeorgreentea:
## What Changed:

## How to test:

## Additional information:


packages/server/core/src/services/index.ts
```diff
@@ -28,6 +30,8 @@ export const services = async (app: Application): Promise<void> => {
   app.configure(document)
   app.configure(projects)
   app.configure(agentImage)
+  app.configure(credentials)
+  if (FEATURE_FLAGS.V2_PLUGINS.includes('slack')) app.configure(slackApi)
```
benbot, line 34: Is there a better place for us to configure different services?

Otherwise we'd need this hard coded dependencies on every plugin's services.

benbot (changes requested):

packages/server/core/src/services/index.ts
coffeeorgreentea: We need an abstraction for the new plug-ins to be able to add services like the old one

michaelsharpe: We can haver the plugin provider a service and load them dynamically from the plugin manager here.
Kind of like how we did with the old one.

michaelsharpe: I would just make an array on the base plugin class for services and add it in there.  And then we can make a helper method called "registerServices" on the main pluginManager with takes in the feathers app and runs over every loaded plugin and registers its services.

michaelsharpe (commented):
So looking at this the slack api class we don't actually need it to be a feathers service.  It can just be a standalone class dependency injected into the plugin without the need for feathers.

coffeeorgreentea:
closed. split into multiple branches

PR #1448, opened 2023-12-27
re-organized node picker menu

FabriceIRANKUNDA:
## What Changed:

Improved node picker

## How to test:

## Additional information:


FabriceIRANKUNDA:
https://github.com/Oneirocom/Magick/assets/58092199/33ac7fe4-9d9f-4ce0-b1ed-02c484a141a1

michaelsharpe:
@FabriceIRANKUNDA this looks great!

One change I would make is once someone starts typing a search it collapses from being menus into being a list for better visibility into search results, and keep the tab autocomplete and arrow select, and the original naming display with the /. 

So normal exploration is with the nested menus (would be nice if these could also be navigated with arrow keys) and search is the flat list.

michaelsharpe:
@FabriceIRANKUNDA Lets get this merged in.  Could you post one final video with the added functionality with search and tab autocomplete?

FabriceIRANKUNDA:
Uploading Screen Recording 2024-01-09 at 12.44.00.mov…

PR #1453, opened 2023-12-31
LiteLLM

michaelsharpe:
## What Changed:

- LiteLLM added with JsPyBridge
- Adds generate text node
- Adds streamMessage node for completion streaming
- chat window in spell composer no supports token streaming

## How to test:

## Additional information:


michaelsharpe:
There are still a few more things we need to do with this:

- improve error handling when LiteLLM throws errors and surface those errors into the IDE
- Maybe merge streaming and non streaming mode so streaming is always happening, and at the end we just return the complete text and trigger done
- Add a <<done>> token when streaming finishes as well
- properly document requirements for spinning up the repo with Python stuff in the Readme
- Ensure this deploys to heroku properly
- Set up API keys in secrets manager for display to user
- Filter choices of models based on what API keys have been provided.

HaruHunab1320 (approved):

HaruHunab1320 (approved):

PR #1458, opened 2024-01-08
Graph events service

FabriceIRANKUNDA:
## What Changed:

## How to test:

## Additional information:


plugins/core/src/lib/nodes/events/messageEvent.ts
```diff
@@ -18,6 +19,15 @@ const messageEventConfig = {
...
     commit('flow')
+
+    saveGraphEvent({
+      sender: event.sender,
+      agentId: event.agentId,
+      connector: 'onMessage',
+      connectorData: JSON.stringify(event.data),
+      content: event.content,
+      eventType: 'Message',
+    })
```
michaelsharpe, line 30: I think we should add this into the BasePlugin class, where the event is processed by the plugin. Right in here:

```typescript
 mapEventsToEventBus() {
    this.events.forEach(event => {
      this.eventEmitter.on(event.eventName, async payload => {
        payload.eventName = event.eventName
        this.logger.debug(
          `Sending event ${event.eventName} to central message bus`
        )
        this.centralEventBus.emit(this.eventQueueName, payload)
        // await this.eventQueue.addJob(event.eventName, payload)
      })
    })
  }
  ```
  
  I also think we might want to store much of the event in a JSONB format, especially since we can query by JSONB properties.  The current event shape looks like the follow.  We don't need all of it, as some of it is mostly for runtime, but some may be important, especially since not all events coming in will have their data just on "content" (I don't think):
  
  ```typescript
  export type EventPayload<
  T = Record<string, unknown>,
  Y = Record<string, unknown>
> = {
  connector: string
  eventName: string
  status: 'success' | 'error' | 'pending' | 'unknown'
  content: string
  sender: string
  observer: string
  client: string
  channel: string
  plugin: string
  agentId: string
  isPlaytest?: boolean
  spellId?: string
  // entities: any[]
  channelType: string
  rawData: unknown
  timestamp: string
  stateKey?: string
  runInfo?: {
    spellId: string
  }
  data: T
  metadata: Y
}
```

michaelsharpe: Maybe a good mid ground is to keep all your hard coded event properties, and then add a JSONB field for the full event.

We should also consider embedding the event contents automatically so they can always be queried by a user. We can do that in a follow up feature push.

michaelsharpe (changes requested):
Overall looks great.  Just the modifications for a JSONB storage of the full event, as well as moving the recording into the BasePlugin class.  

Might be good to make an abstracted GraphEvents service class as well and then we can use that in other places.

